PR #2, opened 2017-12-09
Feat decode wasm

xtuc:


test/compiler/parsing/fixtures/wasm/decode/expected.json
```diff
@@ -0,0 +1,52 @@
...
+          "type": "ModuleExport",
+          "name": "add",
+          "descr": {
+            "type": "func",
+            "id": "id?????"
+          }
+        },
+        {
+          "type": "Func",
+          "id": "id????",
```
xtuc: Define a naming convention for both `id?????`, index is safe

PR #11, opened 2017-12-21
Implement f32.max, f64.max, f32.min and f64.min

maurobringolf:
This is just a start. I think correct handling of `NaN` payloads would require more work than just `Math.min` and `Math.max`. I haven't fully looked into it though. Here are the relevant spec parts:

https://webassembly.github.io/spec/core/exec/numerics.html#op-fmin
https://webassembly.github.io/spec/core/syntax/values.html#syntax-payload
https://webassembly.github.io/spec/core/exec/numerics.html#nan-propagation

### DIFF
--- a/src/interpreter/kernel/exec.js
+++ b/src/interpreter/kernel/exec.js
@@ -786,6 +786,67 @@ export function executeStackFrame(frame: StackFrame, depth: number = 0): any {
       break;
     }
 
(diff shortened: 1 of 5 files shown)


xtuc:
Refs https://github.com/xtuc/js-webassembly-interpreter/issues/1

PR #13, opened 2017-12-22
Refactor use number and ident

xtuc:
Closes https://github.com/xtuc/js-webassembly-interpreter/issues/12

cc @maurobringolf


xtuc:
- `"x"` a name node
- a an identifier node
- 1 is a numberLiteral node

PR #16, opened 2017-12-23
Implemented f32.copysign and f64.copysign

maurobringolf:
Simple implementation that's not spec compliant but a first draft that should get the basics working.


maurobringolf:
What split would you propose?

xtuc:
The file is already divided in some sections:
- Control Instructions
- Administrative Instructions
- Memory Instructions
- Numeric Instructions

I would split it into multiple files based on that, does it make sense to you?

maurobringolf:
That definitely makes sense, yep

PR #17, opened 2017-12-23
First shot at f32.abs

maurobringolf:
I add an extra file to cover unary operators. What do you think about that?


xtuc:
Sorry for the conflict. The code is now in https://github.com/xtuc/js-webassembly-interpreter/blob/master/src/interpreter/kernel/exec/numeric-instructions.js.

maurobringolf:
Is replaced with #20

PR #18, opened 2017-12-23
Refactor kernel exec

xtuc:


src/interpreter/kernel/exec/index.js
```diff
@@ -0,0 +1,210 @@
...
+      pushResult,
+      setLocalByIndex,
+      isZero,
+      getLocalByIndex,
+
+      createChildStackFrame,
+      executeStackFrame,
+    };
+
+    let res;
```
xtuc: I don't like that but it's temporary.

src/interpreter/kernel/exec/index.js
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        if (isTrapped(res)) {
+          return res;
+        }
+
+      };
+    }
+
+    const evaluate = createInstructionEvaluator(Object.assign({},
```
xtuc: This is currently not efficient but i'm using the frameutils object in closure. I will change it in this PR.

PR #19, opened 2017-12-23
bench infra

xtuc:


maurobringolf:
I tried running `make bench` but didnt get any results. Also after re-installing everything, no output and empty `benchmark/result` file.

xtuc:
Currently the results file isn't used. 

This is strange there's no ouput in the CI either. I suspect the WASM binary to be malformed because I doesn't seem to work on Node native WebAssembly. 

I will work on this later today.

Issue #22, opened 2017-12-25
i64 arithmetic on JS?


xtuc:
WebAssembly has a 64 bits float and integer type.

Even if we won't have interactions with >53bits integers between JavaScript code and the WebAssembly runtime (#21).

Our interpreter is still written in JavaScript and we need to handle it somehow, here's a few solutions:
- Emulate 64 bits arithmetic over 32 bits (what asm.js does)
    - We probably need a AST pass to tranform all i64 into i32.
- Compile your program with the 32 bits mode? Which adds a restriction to use our project.

---

asm.js does it already, here's are a few notes that I found:

- Unity in particular mentioned that emulation of 64-bit arithmetic in asm.js is really really slow

maurobringolf:
Since JS uses IEEE double precision floating point numbers (64bit), f32 and f64 are no problem. Short integers i32 are no problem either, since they are representable with double precision doubles. The only type that requires extra handling is i64.

alfonsogarciacaro:
Maybe long.js is an option? That's what Fable uses. https://github.com/dcodeIO/long.js

xtuc:
Thanks @alfonsogarciacaro! That is looking good. Do you know how it affects Fable's performance?

I'm wondering if we could only apply this to numbers > 53bits, so that < 53bits number doesn't suffer from the overhead.

alfonsogarciacaro:
We haven't measured but I guess we could run a bunch of operations with i32 and i64 numbers and see how they compare ([example](https://twitter.com/FableCompiler/status/890488515926122496)).

About discriminating numbers, it may be a bit difficult when you need to operate on them. For example, adding two <53 bits numbers that will result in a >53 bits number, etc.

maurobringolf:
Im with @alfonsogarciacaro, I think the necessary logic to detect the need for >53 bit would produce more overhead and potential bugs than value. Performance will also be more predictable if all i64 values are affected the same way.

xtuc:
Ok, that makes sense.

maurobringolf:
Maybe this is too simplistic, but I just thought about this: For all nodes that hold `i64` values we store two values `lower` and `upper` and define its value as `lower + upper * 2^32`? Because these values are only internal to the wasm computation and cannot be passed in or out to host environment ( #21 ) we never need to represent the complete number in the interpreter. Instead of `+` we use a custom `add64` function that produces two values representing upper and lower.

[...]

Issue #21, opened 2017-12-25
Handle i64 arguments/results in host function


xtuc:
According to the [spec](https://github.com/WebAssembly/design/blob/master/JS.md):

> If the signature contains an i64 (as argument or result), the host function immediately throws a TypeError when called.

The reason is that you can't represent > 53 bits integers in JS where WebAssembly allows up to 64 bits.

PR #23, opened 2017-12-26
Add failing test for i64 arithmetic

xtuc:
Refs #22 

:dancer:

### DIFF
--- a/test/interpreter/fixtures/64bits-arithmetic/module.wast
+++ b/test/interpreter/fixtures/64bits-arithmetic/module.wast
@@ -1,0 +1,10 @@
+(module
+  (func $test (result i32) (
+    (i64.const 9007199254740992)
+    (i64.const 1)
+    (i64.add)
+    (i64.const 9007199254740992)
+    (i64.sub)
+  ))
+  (export "test" (func $test))
+)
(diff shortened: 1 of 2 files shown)


test/interpreter/fixtures/64bits-arithmetic/module.wast
```diff
@@ -0,0 +1,10 @@
+(module
+  (func $test (result i32) (
```
xtuc: Note that I could use i64 in the signature because of #21

ColinEberhardt:
Unfortunately the wast module in this PR doesn't compile:

```
$ wat2wasm test/interpreter/fixtures/64bit-arithmetic/module.wast
test/interpreter/fixtures/64bit-arithmetic/module.wast:7:6: error: type mismatch in implicit return, expected i32 but got i64.
    (i64.sub)
```

xtuc:
Well that makes sense to me:
- the wast is still using an invalid syntax, which I refactored here  https://github.com/xtuc/js-webassembly-interpreter/pull/52
- as mentioned in the error, the function returns an i32 but an i64 was left on the stack. We could implement the conversions to make it work https://github.com/xtuc/js-webassembly-interpreter/issues/63

Sorry about that. We can also add stricter validations like wabt does to avoid that in the future.

ColinEberhardt:
No worries.

I’ve implemented a very crude prototype of i64, that uses long.js. I also came up with the idea of using a cast to to verify in the tests. 

I’ll tidy up a bit and create a PR.

xtuc:
Closing this since it was basically included in https://github.com/xtuc/js-webassembly-interpreter/pull/79

PR #30, opened 2017-12-27
WATF Number parsing

maurobringolf:
This does not cover everything by far, but I think its a fair starting point where we can make separate issues for the remaining cases.


src/compiler/parsing/watf/number-literals.js
```diff
@@ -0,0 +1,35 @@
...
+export function parse64F( sourceString: string ): number {
+  return parseFloat(sourceString);
+}
+
+export function parse32I( sourceString: string ): number {
+  let value = 0;
+  if (isHexLiteral(sourceString)) {
+    value = ~~parseInt(sourceString, 16);
+  } else if (isDecimalExponentLiteral(sourceString)) {
+
```
xtuc: Is this empty? Could we throw an error here, like: unsupported value...

test/compiler/parsing/fixtures/watf/number-literals/f32/decimal-negative-integer/actual.wast
```diff
@@ -0,0 +1,5 @@
+(module
+  (func (
+    (f32.const 33)
```
xtuc: This should probably be a negative number? According to the name of the test.

test/compiler/parsing/fixtures/watf/number-literals/f32/decimal-positive-float/actual.wast
```diff
@@ -0,0 +1,5 @@
+(module
+  (func (
+    (f32.const -98989898)
```
xtuc: And this should probably be position according the its name.

test/compiler/parsing/fixtures/watf/number-literals/f32/decimal-positive-integer/actual.wast
```diff
@@ -0,0 +1,5 @@
+(module
+  (func (
+    (f32.const 0.3)
```
xtuc: It seems that float and int tests should be swapped.

test/compiler/parsing/fixtures/watf/number-literals/i32/hexadecimal-full/expected.json
```diff
@@ -0,0 +1,30 @@
...
+          "result": null,
+          "body": [
+            {
+              "type": "Instr",
+              "id": "const",
+              "object": "i32",
+              "args": [
+                {
+                  "type": "NumberLiteral",
+                  "value": -1
```
xtuc: `0xFFFFFFFF` should fit in an i32?

src/compiler/parsing/watf/number-literals.js
maurobringolf: This is a TODO for literals of the form `234e-3` with decimal exponents. I'll throw an error here and will create issues for all cases that Im aware of that are missing 👍

test/compiler/parsing/fixtures/watf/number-literals/i32/hexadecimal-full/expected.json
xtuc: `0xFFFFFFFF | 0 === -1` lgtm then

PR #38, opened 2017-12-27
feat loop and if instructions exec

xtuc:
Following #28


xtuc:
Merging this for now, probably not every cases are handled but #47 is more important.

PR #42, opened 2017-12-28
Added fp hex notation article link under 'Notes'

maurobringolf:
I added it to *Notes* for now. If there will be more material we can put them into a separate section.


xtuc:
Will also be on the website as soon as I update it.

maurobringolf:
Cool thanks 😃

Issue #1, opened 2017-12-07
Implement all instructions


xtuc:
- [x] `unreachable`
- [x] `nop`
- [x] `block`
- [ ] `loop`
- [x] `if`
- [x] `else`
- [x] `end`
- [ ] `br`
- [ ] `br_if`
- [ ] `brtable`
- [ ] `return`
- [x] `call`
- [ ] `callindirect`
- [x] `drop`
- [ ] `select`
- [x] `getlocal`
- [x] `setlocal`
- [ ] `teelocal`
- [ ] `getglobal`
- [ ] `setglobal`
- [ ] `i32.load`
- [ ] `i64.load`
- [ ] `f32.load`
- [ ] `f64.load`
- [ ] `i32.load8_s`
- [ ] `i32.load8_u`
- [ ] `i32.load16_s`
- [ ] `i32.load16_u`
- [ ] `i64.load8_s`
- [ ] `i64.load8_u`
- [ ] `i64.load16_s`
- [ ] `i64.load16_u`
- [ ] `i64.load32_s`
- [ ] `i64.load32_u`
- [ ] `i32.store`
- [ ] `i64.store`
- [ ] `f32.store`
- [ ] `f64.store`
- [ ] `i32.store8`
- [ ] `i32.store16`
- [ ] `i64.store8`
- [ ] `i64.store16`
- [ ] `i64.store32`
- [ ] `currentmemory`
- [ ] `growmemory`
- [x] `i32.const`
- [x] `i64.const`
- [x] `f32.const`
- [x] `f64.const`
- [ ] `i32.eqz`
- [ ] `i32.eq`
- [ ] `i32.ne`
- [ ] `i32.lt_s`
- [ ] `i32.lt_u`
- [ ] `i32.gt_s`
- [ ] `i32.gt_u`
- [ ] `i32.le_s`
- [ ] `i32.le_u`
- [ ] `i32.ge_s`
- [ ] `i32.ge_u`
- [ ] `i64.eqz`
- [ ] `i64.eq`
- [ ] `i64.ne`
- [ ] `i64.lt_s`
- [ ] `i64.lt_u`
- [ ] `i64.gt_s`
- [ ] `i64.gt_u`
- [ ] `i64.le_s`
- [ ] `i64.le_u`
- [ ] `i64.ge_s`
- [ ] `i64.ge_u`
- [ ] `f32.eq`
- [ ] `f32.ne`
- [ ] `f32.lt`
[...]

PR #44, opened 2017-12-28
Implement simple negate operation

maurobringolf:
#1

### DIFF
--- a/src/interpreter/kernel/exec.js
+++ b/src/interpreter/kernel/exec.js
@@ -994,6 +994,37 @@ export function executeStackFrame(frame: StackFrame, depth: number = 0): any {
       break;
     }
 
+    case 'neg': {
+
+      switch (instruction.object) {
+
+      case 'f32': {
+        const c = pop1('f32');
+
+        pushResult(
+          unopf32(c, 'neg')
+        );
+
+        break;
+      }
+
+      case 'f64': {
+        const c = pop1('f64');
+
+        pushResult(
+          unopf64(c, 'neg')
+        );
+
+        break;
+      }
+
+      default:
+        throw new RuntimeError('Unsupported operation ' + instruction.id + ' on ' + instruction.object);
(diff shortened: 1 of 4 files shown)


src/interpreter/kernel/exec.js
```diff
@@ -991,6 +991,37 @@ export function executeStackFrame(frame: StackFrame, depth: number = 0): any {
         throw new RuntimeError('Unsupported operation ' + instruction.id + ' on ' + instruction.object);
       }
 
+      break;
```
xtuc: Btw do you know if there a rule to ensure the break in ESLint? Sometimes I forget it and loose a lot of time trying to debug it.

test/interpreter/fixtures/fneg/exec.tjs
```diff
@@ -0,0 +1,36 @@
...
+
+  assert.typeOf(res, 'number');
+  assert.equal(res, -0.5);
+});
+
+it('should negate negative input', () => {
+  const res = m.exports.neg(-12341234);
+
+  assert.typeOf(res, 'number');
+  assert.equal(res, 12341234);
```
xtuc: The result type should be a float?

maurobringolf: Not sure what you mean, integers are floats?

src/interpreter/kernel/exec.js
maurobringolf: I don't think so https://github.com/eslint/eslint/issues/7697

test/interpreter/fixtures/fneg/exec.tjs
xtuc: I expected 12341234.0 but you are right.

maurobringolf: Ah yeah these are the same thing in JS

PR #50, opened 2018-01-02
Refactor watf and wast syntax

xtuc:
Closes https://github.com/xtuc/js-webassembly-interpreter/issues/47

## Changes
- Add Babel's code frame when an error occurs
- Allow top-level block instructions (instead of wrapping into a func/module)
- Simplify WAST constructions in tests (mostly removing top-level module when possible)


src/compiler/parsing/watf/grammar.js
```diff
@@ -2,9 +2,15 @@
 
 const {tokens, keywords} = require('./tokenizer');
 const t = require('../../AST');
+const {codeFrameColumns} = require('@babel/code-frame');
```
xtuc: This should probably be a prod dep, I'm not sure yet if we should include it in prod.

xtuc:
Merging this now to avoid an unreviewable PR.

Issue #32, opened 2017-12-27
Floating point hexadecimal format for number literals


maurobringolf:
https://github.com/WebAssembly/spec/blob/master/test/core/float_literals.wast#L95

maurobringolf:
Here is the current deal https://github.com/maurobringolf/webassembly-floating-point-hex-parser

maurobringolf:
I think we can make new issues for uncovered cases. I'm closing this.

PR #54, opened 2018-01-04
Floating point hexadecimals

maurobringolf:
This should add support for the number literal format request in #32 . Installed https://github.com/maurobringolf/webassembly-floating-point-hex-parser, extensive testing happens in that repo. This PR just makes a couple of changes to the tokenizer. I think it's not correct in all cases yet but we can iterate on that.


package.json
```diff
@@ -46,5 +46,8 @@
     "wasmdump": "./lib/bin/wasmdump.js",
     "wasmast": "./lib/bin/wasmast.js",
     "wasmrun": "./lib/bin/wasmrun.js"
+  },
+  "dependencies": {
+    "webassembly-floating-point-hex-parser": "^0.1.2"
```
xtuc: FYI https://github.com/xtuc/js-webassembly-interpreter/commit/575e72db61b36678ab6dde0433518d54b64ffc1d I had to remove the "no dep" in the README and could we fix the version?

xtuc:
You're actually violating your own licence :smile:, could you include the content of your licence in the file https://github.com/xtuc/js-webassembly-interpreter/blob/master/docs/thirdparty-licenses.txt? Thanks!

src/compiler/parsing/watf/tokenizer.js
```diff
@@ -133,7 +133,7 @@ function tokenize(input: string) {
     }
 
     const NUMBERS = /[0-9|.|_]/;
-    const HEX_NUMBERS = /[0-9|A-F|a-f|_|.]/;
+    const HEX_NUMBERS = /[0-9|A-F|a-f|_|.|p|P|-]/;
```
xtuc: How do you parse `nan:0x40000`? Could we add a test for this case?

maurobringolf: These are not included in here, I am working on it in a separate branch:

https://github.com/xtuc/js-webassembly-interpreter/tree/nan-payloads

I planned to fix all merge conflicts before PRing from the nan-payloads branch and do this separately.

xtuc: ok

xtuc:
Thanks for the changes, we can merge this. Btw i'll upgrade your tests once https://github.com/xtuc/js-webassembly-interpreter/pull/52 is merged

PR #58, opened 2018-01-06
Add note for kernel/exec structure

xtuc:
Following the discussion in https://github.com/xtuc/js-webassembly-interpreter/issues/56#issuecomment-355736921

Let me please know what do you think about this? Is is clear enough?

cc @ColinEberhardt


ColinEberhardt:
Looks good to me!

Issue #56, opened 2018-01-06
Question on programming style


ColinEberhardt:
Looking at the implementation of `add`, there is quite a bit of repetition:

    case 'add': {

      switch (instruction.object) {

      case 'i32': {
        const [c1, c2] = pop2('i32', 'i32');

        pushResult(
          binopi32(c2, c1, '+')
        );

        break;
      }

      case 'i64': {
        const [c1, c2] = pop2('i64', 'i64');

        pushResult(
          binopi64(c2, c1, '+')
        );

        break;
      }

      case 'f32': {
        const [c1, c2] = pop2('f32', 'f32');

        pushResult(
          binopf32(c2, c1, '+')
        );

        break;
      }

      case 'f64': {
        const [c1, c2] = pop2('f64', 'f64');

        pushResult(
          binopf64(c2, c1, '+')
        );

        break;
      }

      default:
        throw new RuntimeError(
          'Unsupported operation ' + instruction.id + ' on ' + instruction.object
        );

      }

      break;
    }

The following refactor gives the same functionality (ignoring the unsupported operation path for now):

    case 'add': {

      const binopComputer = {
        'i32': binopi32,
        'i64': binopi64,
        'f32': binopf32,
        'f64': binopf64,
      };

      const [c1, c2] = pop2(instruction.object, instruction.object);

      pushResult(
        binopComputer[instruction.object](c2, c1, '+')
      );

      break;

    }

Is a more 'long hand' approach being used for reasons of performance? Or is it worth exploring refactoring further?

maurobringolf:
I doubt that this refactoring has a measurable performance impact, so from my perspective this refactoring would definitely be a win 👍

ColinEberhardt:
Cool, I’ll propose some ‘light touch’ refactoring in a PR before I do more work on the binary ops.

xtuc:
It's definitely more readable.

FYI I tried to split the kernel/exec file into a more convenient way to declare operations, the PR is here https://github.com/xtuc/js-webassembly-interpreter/pull/18. Unfortunately it became much more slower (https://github.com/xtuc/js-webassembly-interpreter/commit/1fdad4f43a073a2b28cfae58ce9a9619bd67e61f), and I had to revert it.

We have a make target if you want to run the benchmark, I believe it will be fine because it's not the whole interpreter like I tried.

ColinEberhardt:
[...]

PR #67, opened 2018-01-06
interpreter exec refactor

ColinEberhardt:
Here is a proposed refactor ...

1. the binop computer now uses the instruction names, rather than JavaScript 'signs', i.e. '+' becomes 'add', removing a further look-up
2. kernel/exec has been refactored so that all binary operations share the same 'case' block.

I committed by own benchmark result before the refactor. Here is the mean time before and after:

mean 474.9725275000 ns
mean 481.3720141000 ns

If everyone is happy with this, I'll look at a similar change for unary operations.

(before merging, I'll remove the benchmark commit, and tidy up a few comments)

relates to #56

### DIFF
--- a/src/interpreter/kernel/exec.js
+++ b/src/interpreter/kernel/exec.js
@@ -724,65 +724,38 @@ export function executeStackFrame(frame: StackFrame, depth: number = 0): any {
     }
 
     /**
-     * Numeric Instructions
-     *
-     * https://webassembly.github.io/spec/exec/instructions.html#numeric-instructions
+     * Unary operations
      */
-
-    case 'abs': {
-
-      switch (instruction.object) {
-
-      case 'f32': {
-        const c = pop1('f32');
-
-        pushResult(
-          unopf32(c, 'abs')
-        );
-
-        break;
-      }
-
-      default:
-        throw new RuntimeError('Unsupported operation ' + instruction.id + ' on ' + instruction.object);
-      }
-
-      break;
-    }
-
-
+    case 'abs':
     case 'neg': {
-
+      let unopFn;
       switch (instruction.object) {
-
-      case 'f32': {
-        const c = pop1('f32');
-
-        pushResult(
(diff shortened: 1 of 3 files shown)


ColinEberhardt:
Travis has fallen over again, during `npm install`

maurobringolf:
👍  For using the `instruction.id` all the way through, its more consistent like this.

src/interpreter/kernel/exec.js
```diff
@@ -678,328 +678,58 @@ export function executeStackFrame(frame: StackFrame, depth: number = 0): any {
...
+    case 'sub':
+    case 'div_s':
+    case 'div_u':
+    case 'div':
+    case 'min':
+    case 'max':
+    case 'copysign':
+    case 'or':
+    case 'xor': {
+      let binopComputer;
```
xtuc: Sounds a bit confusing IMO. What about binopFn or binopCalculator?

benchmark/add/results
```diff
@@ -1,13 +1,13 @@
...
-mean 21.7151342000 ns
+total 255.1195230000 ms
+mean 25.5119523000 ns
 
 Interations: 10000000
-Date: 2017-12-29
+Date: 2018-1-6
 V8 version: 6.2.414.46-node.15
-Interpreter version: 0.0.14
+Interpreter version: 0.0.15
```
xtuc: We can totally merge this file, but I should probably run it from my machine to keep it consistent, your native result seems a bit slower.

ColinEberhardt:
> Sounds a bit confusing IMO. What about binopFn or binopCalculator?

OK, I've gone for `binopFn`. I've also rebased out my benchmark results. Should be good to go now.

xtuc:
Here are the results: https://github.com/xtuc/js-webassembly-interpreter/commit/32a5fd2158081a1416d505d25160e3f26ac782a9 :+1:

Issue #45, opened 2017-12-28
Capture lost information of number values in JS floats


maurobringolf:
Since we are interpreting in JS it makes sense to use JS floats for all numbers and their operations wherever possible. However, I think there are a few problems with that:

* **NaN**: There are many different `NaN` representations containing a payload. I don't think that there is a way to access the payload in JS (or that it is propagated through operations at all). But the WebAssembly spec says that the payloads should propagate through operations according to well-defined rules https://webassembly.github.io/spec/core/exec/numerics.html#aux-nans. I think this is important, because the bits of a `NaN` could be reinterpreted as an `i64` integer for example. The same applies to the sign bit I think.

* **+-0**:  It is possible to distinguish `+0` and `-0` in JS but not with `===`. That is why test cases with these values are currently useless, because the testing framework in JS treats them as the same. Again, I think we need to store some extra information about the sign.

I will try to come up with some simple examples and run them through the reference interpreter to make the issue more clear. We dont have enough implemented yet to show the `NaN` issues, but here is a sample test case with the `+-0` issue https://github.com/xtuc/js-webassembly-interpreter/blob/master/test/interpreter/kernel/exec/numeric-instructions.js#L383-L398.

xtuc:
It's clear enough to me. 

I would box every numbers in JS and add this informations. 

Btw I recommend you testing with https://mbebenita.github.io/WasmExplorer/

maurobringolf:
#49 Fixes the test issue with `+-0`. I think its enough to store the extra information for `NaN`s. My idea would be to make a `maybeNumber` type that looks something like this (pseudocode):

```javascript
{
  isNaN: false|true,
  value: number
}
```

The payload of a `NaN` is a positive integer, so we can store the sign with it. If the value is negative, then its absolute value is the payload and the `NaN` has a negative sign.

We'd have to change all existing floating point operations though, such that they work with this kind of inputs instead of numbers.

What do you think @xtuc ? I'm happy to do it, but maybe there is a simpler way.

xtuc:
Lgtm. We need to make sure to not use Number.isNaN anymore. 

I think that we can store the sign too. We could use it for faster comparaison or for the i64 (2x i32) type.

PR #70, opened 2018-01-07
[do not merge] i64 maths quick and dirty prototype

ColinEberhardt:
This is a quick prototype to explore i64 maths as per #22 

(needs rebasing on top of #67)

The key points are:

1. Uses long.js for parsing strings to i64 and for maths operations
2. Introduces a new type to the AST, `LongLiteral`, where the value has `upper` and `lower` components.
3. At runtime, i64 values are stored as instances of `Long`
4. Testing uses a `f64.reinterpret/i64` to cast to floating point. I'd propose treating the implementation of long.js as a 'trusted unit', which means this project would only need lightweight integration tests.
5. I've commented out a bunch of tests just to keep the 'noise' down from `npm test` ;-)

I'm most interested in feedback on points 1, 2 & 3 above.

Also, there are two tests that are failing, both have some form of 'magic number'. I'll explore further.

```
  215 passing (162ms)
  2 failing

  1) compiler
       WATF parsing
         test/compiler/parsing/fixtures/watf/block/empty/actual.wast:
     Error: 
- Expected
+ Received

  {
    "type": "Program",
    "body": [
      {
        "type": "BlockInstruction",
        "id": "block",
        "label": {
          "type": "Identifier",
-         "name": "block_14"
+         "name": "block_13"
        },
        "instr": []
      }
    ]
  }
      at check (test/compiler/parsing/index.js:47:19)
      at Object.parseWATF (lib/index.js:73:5)
      at Promise (test/compiler/parsing/index.js:60:16)
      at new Promise (<anonymous>)
      at Context.it (test/compiler/parsing/index.js:28:23)

  2) interpreter
       watf
         test/interpreter/fixtures/call-exported-func-with-mismatching-arity/module.wast
           should throw an error for invalid arity:

      AssertionError: expected [Function: fn] to throw error including 'Function 1 called with 4 arguments but 2 expected' but got 'Function 2 called with 4 arguments but 2 expected'
      + expected - actual

      -Function 2 called with 4 arguments but 2 expected
      +Function 1 called with 4 arguments but 2 expected
      
      at Context.it (test/interpreter/fixtures/call-exported-func-with-mismatching-arity/module.wast:6:10)
```


src/compiler/AST/index.js
```diff
@@ -166,7 +166,7 @@ export function numberLiteral(
   }
 
   return {
-    type: 'NumberLiteral',
+    type: type === 'i64' ? 'LongLiteral' : 'NumberLiteral',
```
xtuc: We actually discussed about that with @maurobringolf offline (next time we will make an issue, sorry).

I saw that `i8` and `i16` were also around (unspecified now tho) and proposed to introduce more granular types:

```js
// using flow syntax

interface NumberLiteral {
  type: 'NumberLiteral';
  value: any;
}

type i8 = NumberLiteral & { value: number; }
type i16 = NumberLiteral & { value: number; }
type i32 = NumberLiteral & { value: number; }
type i64 = NumberLiteral & { hi: i32; low: i32; } // because JS can't handle 64 bits int
```

And the according AST builders.

Do we agree on this? cc @maurobringolf

src/interpreter/kernel/exec.js
```diff
@@ -1089,6 +1089,15 @@ export function executeStackFrame(frame: StackFrame, depth: number = 0): any {
 
       break;
     }
+
+    case 'reinterpret/i64': {
+      
+      const c1 = pop1('i64');
+      const res = f64.createValue(c1.value.toNumber());
```
xtuc: There are probably some cases where `c1.value.toNumber()` returns an invalid value?

Note that currently `pushResult` can push `undefined` on the stack (I have duplicated code all over the kernel/exec to handle this case, we could maybe include it directly in pushResult).

src/interpreter/kernel/instruction/binop.js
```diff
@@ -58,12 +59,41 @@ function binop(
...
+  c1: StackLocal,
+  c2: StackLocal,
+  sign: Sign,
+  createValue: (Long) => StackLocal
+): StackLocal {
+  
+  let res;
+
+  switch (sign) {
+  // https://webassembly.github.io/spec/exec/numerics.html#op-iadd
```
xtuc: Btw this link changed: https://webassembly.github.io/spec/core/exec/numerics.html#op-iadd

src/interpreter/runtime/values/i64.js
```diff
@@ -1,14 +1,12 @@
...
 const type = 'i64';
 
-export function createValue(value: number): StackLocal {
-  value = (value | 0) % Math.pow(2, bits);
-
+export function createValue(value: Long): StackLocal {
   return {
     type,
-    value,
+    value: new Long(value.lower, value.upper)
```
xtuc: It's not a big deal but I like trailing commas (because you can duplicate a line more easily). Might be worth adding an ESLint/Prettier to keep our code consistent.

src/types/AST.js
```diff
@@ -2,7 +2,7 @@
 
 type Valtype = 'i32' | 'i64' | 'f32' | 'f64' | 'label';
 type ExportDescr = 'func' | 'table' | 'memory' | 'global';
-type Index = NumberLiteral | Identifier;
+type Index = NumberLiteral | LongLiteral | Identifier;
```
xtuc: The `Index` concept might be to vague here.

For example, the CallInstruction take an index because:
- In WAST `(call $foo)` is an identifier
- In WASM/WATF `(call 0)` is an NumberLiteral (and by spec an unsigned i32)

I will try to clarify it.

test/interpreter/kernel/exec/numeric-instructions.js
```diff
@@ -102,92 +102,10 @@ describe('kernel exec - numeric instructions', () => {
 
     /**
      * Integer 64 bits
+     * 
+     * TODO: put these tests back in place!
```
xtuc: These tests are actually not feasible in JS. Might be worth investigating another way to test (stack serialization comparison, I think I saw ChakraCore doing it so)

xtuc:
Don't worry about the failing "magic number" tests, it's due to https://github.com/xtuc/js-webassembly-interpreter/blob/c8f73a14d64ff3b44016a8394957dad9c1470ae5/src/compiler/parsing/watf/grammar.js#L20-L24

I know it's not the best way to deal with it, you can remove the `expected.json` for now.

src/interpreter/runtime/values/i64.js
ColinEberhardt: definitely +1 to the idea of adding prettier, I was going to ask about what code formatter / linter you are using locally :-)

ColinEberhardt: on a related note, I guess you are using Flow in an ad-hoc fashion at the moment? I'm seeing quite a lot of Flow errors when running `make lint`

xtuc: Yes, that's absolutely my bad, I will have to fix them. For convenience (and to be consistent in my other JS projects) I configured ESLint to call Flow, I only have an integration with ESLint in my editor.

src/types/AST.js
xtuc: Refs https://github.com/xtuc/js-webassembly-interpreter/issues/69 where some Identifier will be transformed to NumberLiterals. The idea is to remove `type Index = Identifier` completely in the future.

src/compiler/AST/index.js
maurobringolf: Yes agree 👍

src/compiler/parsing/watf/number-literals.js
```diff
@@ -29,8 +29,12 @@ export function parse32I( sourceString: string ): number {
...
 }
 
-export function parse64I( sourceString: string ): number {
-  return parseInt(sourceString);
+export function parse64I( sourceString: string ): number | Long {
+  const long = Long.fromString(sourceString);
+  return {
+    upper: long.high,
+    lower: long.low
+  };
```
maurobringolf: I think we might need to use `isHexLiteral()` here to give ` Long.fromString` the correct radix argument?

src/interpreter/kernel/exec.js
maurobringolf: I don't think that `Long.toNumber` does the same thing as `reinterpret/i64`. Long performs rounding -> changes bits to get the most similar value. Reinterpretation does not change the bits. In this case it should take upper and lower stored by Long and interpret these 64 bits as IEEE binary64 floating point number.

* [**Long.toNumber**](https://github.com/dcodeIO/long.js#longtonumber)
* [**reinterpret/i64**](https://webassembly.github.io/spec/core/exec/numerics.html#op-reinterpret)

src/interpreter/runtime/values/i64.js
maurobringolf: 🎉 🎉  Prettier, yeess 🎉 🎉

src/interpreter/kernel/exec.js
ColinEberhardt: Exactly, this was a hack just to get something testable out the other end - it is not a reinterpret cast.

The key things I wanted to discuss via this PR are the use of long.js and the AST format. Many of the other details are ‘hacked’.

I’d be happy to iterate either and create a more appropriate PR if the overall approach is ok?

maurobringolf: Okay nice! Currently we don't have the necessary `NaN` payload support in place to get this completely right anyways (#45). I think Long.js will be an overhead in the long ( 😁  ) run, but its definitely a good starting point as discussed in #22, so Im happy with it.

ColinEberhardt:
TODO:

 - [ ] fully implement binop function for i64 numbers - and unit test
 - [ ] rebase on top of #67
 - [ ] remove reinterprit implementation, re-adding an a future PR
 - [ ] update comments where spec URL has changed
 - [ ] use `high` / `low` as the names of the long components
 - [ ] re-instate the deleted tests

xtuc:
Sounds good!

On my side:
- I can enable prettier and reformat the whole project but that will likely break your PR, ok to you?.
- We could refactor the types before and rebasing your PR with it (at least separate i32 and i64), ok to you?.

ColinEberhardt:
> I can enable prettier and reformat the whole project but that will likely break your PR, ok to you?

No problem, I'd be happy to rebase my changes on top of that. 

> We could refactor the types before and rebasing your PR with it (at least separate i32 and i64 see #70 (comment)), ok to you?

Likewise, please go ahead. I've not used Flow before (I generally use TypeScript instead), so it's probably better if you do that initially.

> or anything else that blocks you?

No, I'm all good. I'll hopefully have something ready in a couple of days.

PR #73, opened 2018-01-07
WAST identifiers to index

xtuc:
Closes https://github.com/xtuc/js-webassembly-interpreter/issues/69


src/compiler/AST/traverse.js
```diff
@@ -1,22 +1,9 @@
...
-  }
-  // $FlowIgnore
-  return output;
-}
-
 function createPath(node: Node): NodePath<Node> {
 
   return {
-    node: cloneNode(node),
+    node,
```
xtuc: Well, I tried to use an immutable tree but didn't succeed. I hope we won't end up in the same state as Babel with unexpected mutations. I think it will be fine since we won't have much transformations like this here.

test/compiler/parsing/fixtures/watf/wast-identifier-to-index/actual.wast
```diff
@@ -0,0 +1,9 @@
+(func (param $c f64) (param $a i32) (param $b i64)
+  (get_local $a)
+  (get_local $b)
+  (get_local $c)
+  (get_local $a)
+
+  (set_local $a (i32.const 1))
+  (tee_local $a (i32.const 1))
+)
```
xtuc: Sorry for unrelated tests modification, this is the interesting one.

Basically replacing `(get_local $c)` with its offset in the locals: `(get_local 0)`.

src/compiler/parsing/watf/index.js
```diff
@@ -2,14 +2,15 @@
...
+const wastIdentifierToIndex = require('../../transform/wast-identifier-to-index');
 
-function parseSource(source: string): Program {
+export function parseSource(source: string): Program {
   const tokens = tokenize(source);
 
   // We pass the source here to show code frames
   const ast = parse(tokens, source);
 
+  wastIdentifierToIndex.transform(ast);
```
xtuc: I decided to add the transformation right after the parsing, it will keep things more consistent. But I had to modify some unrelated tests to make them pass.

xtuc:
Travis is failing right now but it's unrelated, the tests are passing locally.

xtuc:
Adding some docs about this in https://github.com/xtuc/js-webassembly-interpreter/commit/442b3844a677bdc227c677d921a20bdc6d44439c

PR #75, opened 2018-01-07
Fix nested instructions

xtuc:
Closes https://github.com/xtuc/js-webassembly-interpreter/issues/57 and removes some unnecessary complexity.


src/compiler/parsing/watf/grammar.js
```diff
@@ -82,73 +82,6 @@ export function parse(tokensList: Array<Object>, source: string): Program {
       return true;
     }
 
-    function parseMaybeSignature() {
```
xtuc: That's unrelated, it wasn't used anymore :smile:

PR #78, opened 2018-01-08
Prettier

xtuc:
Closes https://github.com/xtuc/js-webassembly-interpreter/issues/77

I'm not a super fan of the pre-commit hook tho


xtuc:
@maurobringolf @ColinEberhardt are you guys happy with this?

You can run `make fix` to prettify your PRs. Note that it is currently conflicting with ESLint, but it's not a big deal since we don't use it really.

ColinEberhardt:
That's great - thanks. Just a couple of questions:

- Regarding the type definitions, do these have to be committed? With TypeScript you just add them to your project dependencies, e.g. `@types/lodash`.
- Do you still need ESLint in the project if prettier is being used to enforce a standard format?

> I'm not a super fan of the pre-commit hook tho

Nor am I - I just use the vscode prettier plugin to format-on-save, and CI to ensure others do the same.

xtuc:
> Regarding the type definitions, do these have to be committed? With TypeScript you just add them to your project dependencies, e.g. @types/lodash.

I don't use `flow-typed` so often, but as far as I can see on the internet, it seems pretty common to actually commit them. I preferred putting them into `./src/types` instead of in the root of the project. The definitions can be update using a make target.

> Do you still need ESLint in the project if prettier is being used to enforce a standard format?

Prettier only ensure code formating, we can still use ESLint for running flow and different linting rules (unused vars, missing returns, etc).

> Nor am I - I just use the vscode prettier plugin to format-on-save, and CI to ensure others do the same.

Happy to hear that! Let me change this here then.

xtuc:
Ok @ColinEberhardt, now basically everything goes through ESLint, I added the linting in the CI (I had to disable flowtype errors for now, I will work on this).

xtuc:
I'll go ahead and merge this, feel free to comment here if you need to change smth.

ColinEberhardt:
One other thought ... does this also lint the .tjs files? (I’ve not tried yet, but can’t recall spotting anything in the PR that includes them in the eslint glob)

xtuc:
Good catch! .tjs are not included. I will add them.

They have custom extension because mocha is in recursive mode and would try to execute them directly otherwise.

PR #79, opened 2018-01-09
feat: implemented i64 operations

ColinEberhardt:
I've still got to rebase and use 'prettier', but thought this worth worth putting up for review now. I think I'll have 30 mins free tomorrow to rebase and test.

Quick summary:

- uses long.js for parsing, operations and the i64 representation at runtime
- implements the WATF parsing logic to match that of i32 (hex, decimal and underscores)
- adds a new `LongLiteral` type to the AST, wth `high` and `low` properties as described by the `LongNumber` interface.
- i64 binary operations are unit tested, but not exhaustively (i.e. no boundary values etc ...), as I am trusting the long.js implementation
- there is no integration-level testing via `test/interpreter/fixtures`. My plan is to add this in a future PR when reinterpret is implemented.

It would be great to get some feedback on the above, and in the meantime I'll rebase and prettier the code!

replaces #70 

Here is the todo list from PR #70 

 - [x] fully implement binop function for i64 numbers - and unit test
 - [x] rebase on top of #67
 - [x] remove reinterprit implementation, re-adding an a future PR
 - [x] update comments where spec URL has changed
 - [x] use `high` / `low` as the names of the long components
 - [x] re-instate the deleted tests
 - [x] fully implement the wat parsing to support the various number formats
  
There are a few points where I copied the existing i32 logic, for my own benefit, there are a few finer details I want to check against the spec. However, for now, I've aimed for consistency.


src/interpreter/kernel/instruction/binop.js
```diff
@@ -9,52 +10,116 @@ const f64 = require('../../runtime/values/f64');
...
+  {value: c1} : StackLocal,
+  {value: c2} : StackLocal,
+  sign: Sign,
+  createValue: (LongNumber) => StackLocal
+): StackLocal {
+  
+  let longResult: LongNumber;
+
+  switch (sign) {
+  // https://webassembly.github.io/spec/exec/numerics.html#op-iadd
```
xtuc: Specs links are broken here, but I can update them all at once (search and replace probably). https://github.com/xtuc/js-webassembly-interpreter/issues/80

src/types/AST.js
```diff
@@ -13,6 +13,11 @@ type NodePath<T> = {
  * AST types
  */
 
+interface LongNumber {
```
xtuc: Could we rename it to `I64Literal` or `i64`?

What do you think of https://github.com/xtuc/js-webassembly-interpreter/pull/70#discussion_r160047149? hi and low as I32 seems a bit overkill tho

test/interpreter/kernel/exec/numeric-instructions.js
```diff
@@ -748,10 +799,15 @@ describe('kernel exec - numeric instructions', () => {
     describe(op.name, () => {
       it('should get the correct result', () => {
 
+        debugger;
         const stackFrame = createStackFrame(op.code, op.args);
         const res = executeStackFrame(stackFrame).value;
 
-        assert.deepEqual(res, op.resEqual);
+        if (res.value instanceof Long) {
+          assert.isTrue(res.value.equals(op.resEqual));
```
xtuc: That's a good news, we can test WebAssembly code involving i64 in JS.

xtuc:
- Prettier is meant to be idempotent, I would recommend you to run `make lint` on the code base and then rebase with the actual PR (where I introduced Prettier), it will reduce the number of conflicts.
- I need to fix the Flowtype checking because we can't rely on it atm - https://github.com/xtuc/js-webassembly-interpreter/issues/81

ColinEberhardt:
I’ve done a quick update, rebasing, applying prettier and also updating the comments. I’ll take a look at our suggestion regarding types in a bit when I have time for a bit of thinking!

src/types/AST.js
ColinEberhardt: > Could we rename it to `I64Literal` or `i64`?

I don't think `I64Literal` makes sense. The current `NumberLiteral` interface describes an AST node type, whereas `LongNumber` describes the representation of a 'long' number as having `high` and `low` properties.

The corresponding literal, that describes an AST node that contains a long value is [further down the same file](https://github.com/xtuc/js-webassembly-interpreter/pull/79/files#diff-f0fcb9b6c38a452cae948d4b45ab85ffR50).

Perhaps `i64` would make sense, although I feel it would look a little out of place with the naming for the other interfaces / types in this file.

> What do you think of #70 (comment)? hi and low as I32 seems a bit overkill tho

This does feel overkill to me to. Unless I'm mistaken, the information these interfaces convey would actually be lost in the AST itself? i.e. it would yield exactly the same output.

src/types/AST.js
```diff
@@ -42,6 +47,11 @@ interface NumberLiteral {
   value: number;
 }
 
+interface LongLiteral {
```
ColinEberhardt: @xtuc perhaps this should be named `LongNumberLiteral`?

i.e. the `LongNumber` interface is intended to capture a concept which is similar to JavaScripts built in `number` type.

src/types/AST.js
xtuc: > I don't think I64Literal makes sense

Yes you are right, I'm :+1: for `LongNumberLiteral` AST and `LongNumber` for the runtime instance. 

Having I64 = [i32, i32] is not very useful from the AST point of view, we just need i64 = [number, number]

ColinEberhardt: done :-) https://github.com/xtuc/js-webassembly-interpreter/pull/79/commits/128ac7973bd3bc0262adc9d9ae62435f73567f1a

ColinEberhardt:
> LGTM, you have done an incredible good job here!

Thanks :-)

PR #84, opened 2018-01-09
Fix flowtype

### LINKED ISSUE #81: Fix flow type checking

xtuc:
Closes #81 

## Breaking change

I fixed something that annoyed me since some time, `Identifier` and `NumberLiteral` have both a `value` field now (instead of one having `name`).


xtuc:
This is becoming to big and requires https://github.com/xtuc/js-webassembly-interpreter/pull/90, i'll merge this now and will iterate on it in the future.

PR #85, opened 2018-01-09
[WIP] Nan payloads

maurobringolf:
I haven't done a lot yet and what I've done is quick and dirty. I'm opening the PR early so you guys can see what I am working on. I will try and rebase it now.


test/compiler/parsing/fixtures/watf/nan/payload/actual.wast
```diff
@@ -0,0 +1,3 @@
+(func (
+  f32.const nan:0x400012
```
xtuc: This works because you only have one instructions here.

Could you reformat it to:
```wat
(func (f32.const nan:0x400012))

;; or

(func (
  f32.const nan:0x400012
))
```

ColinEberhardt:
Looks good to me. Minor points, I'd simply use the property `nan` (rather than `isnan`) in the AST. The 'is' prefix reminds me a little of Hungarian Notation, which I was never keen on.

Also, I'd love to know what the use is of `nan:0x400012`, why would you want a NaN that also has a numeric value?

maurobringolf:
@ColinEberhardt I can use `nan` and `inf` for AST properties, thats fine with me 👍 

I can't say much about why this is useful. It is useful whenever the compiler who generates WebAssembly figures out that it is useful. Not sure how often that happens to be honest.

I should have referenced the relevant issue, my bad ( #45 ). In JS we only have one `NaN` but WebAssembly defines them to be distinct. The payload is the bits of the significand / mantissa (nonzero). That's why we need to store the values and can't use JS `NaN`s (apart from the fact that JSON in the AST doesnt have `NaN` at all).

ColinEberhardt:
Thanks for the clarification - so this isn't just an issue with parsing WAT? You're going to have to store additional information alongside numbers (that represent floats) at runtime? This is similar to the handling of i64 where the Long type (from long.js) is used at runtime - you might have to create your own floating point wrapper that implements the NaN propagation rules you reference in #45

maurobringolf:
Yes exactly. This is not specific to WAT, I just started to work on it in WAT because I already know how to write tests there. The node type is definitely something that needs to be defined, now that we split NumberLiteral into multiple types. Would `FloatLiteral` make sense? cc @xtuc?

> your own floating point wrapper that implements the NaN propagation rules

I planned to implement the propagation logic inside individual instructions. I'm not sure if you are suggesting a different approach

ColinEberhardt:
> I planned to implement the propagation logic inside individual instructions. I'm not sure if you are suggesting a different approach

Inspired by long.js, I think it's worth thinking about alternatives, as I've just commented here:

https://github.com/xtuc/js-webassembly-interpreter/issues/87#issuecomment-356540383

xtuc:
> @ColinEberhardt I can use nan and inf for AST properties, thats fine with me :+1:

Agreed.

> Would FloatLiteral make sense

Yes!

xtuc:
Could you please run `make fix` to make the link pass?

maurobringolf:
Sorry about that. I am trying to change the node type to be `FloatLiteral` now.

maurobringolf:
@xtuc What is the node type of `0` supposed to be in the following?

```wasm
(func (br_if 0))
```

It is a label and not a program value like an i32 or f32. It seems wrong to me that it is a `numberLiteral`. Tests are failing because the default type of the `numberLiteral` constructor is `f64` and labels currently use `numberLiteral` without a type -> they become `floatLiterals` instead. How should I go about this?

xtuc:
According to the spec: `br_if labelidx`, where `labelidx` is an u32 (in general every index is an unsigned 32bits int).

**but** in the WAST syntax `labelidx` could also be an identifier.

To stay as consistent as possible I wrote https://github.com/xtuc/js-webassembly-interpreter/blob/master/docs/contrib/transform/wast-identifier-to-index.md, so it should be an i32 (if not it's a bug!).

maurobringolf:
I see, didn't realize that the name labels were not part of WATF at all. I updated `numberLiteral` to default to `i32`, so all calls should to it for labels should still produce the correct type. I updated the tests where `floatLiterals` are produced.

Do I need to add the type somewhere else? I'm not very familiar with flow yet.

xtuc:
> labels were not part of WATF at all

Yes! **but** they are in WAST. That why our type `Index = NumberLiteral | Identifier`.

Sorry if flow isn't _super super_ usable atm, I need to work on #84

ColinEberhardt:
This reminds me, I've been meaning to ask this for a while ...

Where are the WATF and WAST formats described?

The WebAssembly website gives [an overview](http://webassembly.org/docs/text-format/) of the text format, and describes that it uses S-expressions. It recommends a `.wat` file extension.

It also states:

> Note: The .wast format understood by some of the listed tools is a superset of the .wat format that is intended for writing test scripts. 

Are you using WATF to refer to `.wat` files and WAST to the `.wast` format (as used by the WebAssembly test suite).

xtuc:
To clarify I just wrote https://github.com/xtuc/js-webassembly-interpreter/blob/master/docs/contrib/watf-vs-wast.md.

ColinEberhardt:
> To clarify I just wrote https://github.com/xtuc/js-webassembly-interpreter/blob/master/docs/contrib/watf-vs-wast.md.

Thanks - that clarification is certainly needed!

> We can setup a video chat if you want.

Nice idea, what timezones are you both in? I'm UK-based.

maurobringolf:
Switzerland and @xtuc from Strasbourg, so time zone shouldn't be an issue 👍 Good idea!

xtuc:
Yes, what about tommorow? It's probably easier to talk off, Slack or email? We are already on Babel's slack.

xtuc:
Closes https://github.com/xtuc/js-webassembly-interpreter/issues/94

maurobringolf:
So I rebased and got lots of failing tests.  I pretty much reverted anything from #93 I believe, because this PR implements `nan` and `inf` keywords as `FloatLiteral` nodes, so there is no need for an Identifier (Sorry @xtuc ). This way #94 is automatically avoided I think.

We could switch to unaryExpression nodes to handle signs, but it would require lots of refactoring (all places where we currently have numberLiterals). I think it is easier this way.

I am not sure how this PR relates to #90 . Here `numberLiteral` grew quite large and handles a lot of cases which I don't like. I think the proposed changes in that PR help with this, but I have not looked at it in detail. Probably a good discussion topic for tomorrow.

maurobringolf:
Okay so all existing test cases are parsed correctly now. The number literal parser is not very good yet, but I would like to improve it and its test in a new PR.

I will add some operations on floats before refactoring the AST in the way we discussed this morning (sign field etc). I think the best representation of inf and nan can be determined once we have some operations implemented, then I can argue for or against the current one.

maurobringolf:
## Example implementation of `nan` addition

I implemented f32 addition with correct `nan` payload propagation using a new subtype of `f32`, namely `f32nan`. Also I had to manipulate the test file a bit because we only passed `type` and `value` to the `castIntoStackLocalOfType` function which is not enough to handle `nan` (need the payload **and** some flag to indicate that it is a nan).

Let me know what you think about the subtyping.

## `inf` and `nan` literals**
So I did not change the `inf` and `nan` representation in the AST. They are currently done like this:

```javascript
// inf
{
  type: "f32",
  value: +-1, // Determines the sign
  inf: true
}

// nan
{
  type: "f32",
  value: 1234 // payload, contains the sign
  nan: true
}
```

Currently we don't have `+-0` as number correctly implemented. But I'd rather do this in a separate PR. @xtuc I thought about the sign in infinity and our discussion this morning. I think with plus and minus 1 the code will be simpler, because we can use `node.value * Infinity` to create the value in JS.

xtuc:
@maurobringolf yes ok, storing the sign into the value is fine to me. We could use `Math.sign(node.value) * Infinity` just in case the value will be higher than -1/1 which would be a bug.

maurobringolf:
@xtuc Can we use flow to make a type *sign* that only contains -1 and 1 as possible values? I need to look at flow in more detail. Your suggestion is a nice safety mechanism in any case 👍 

Should I implement more tests and refactor existing cases to use `nan` and `inf` fields? Or how should we proceed with this?

Once we get this right and merge it I will refactor the number literal parser in text format.

xtuc:
> @xtuc Can we use flow to make a type sign that only contains -1 and 1 as possible values? I need to look at flow in more detail. Your suggestion is a nice safety mechanism in any case :+1:

Yup, sounds good. I'm incrementally fixing Flowtype at the moment (I added a few commits in your PR btw).

> Should I implement more tests and refactor existing cases to use nan and inf fields? Or how should we proceed with this?

I mean we should at least have one test with nan and inf fields. I'm not sure to understand what do you mean by refactor existing cases? The new attributes may just be present on some nodes.

test/interpreter/kernel/exec/numeric-instructions.js
```diff
@@ -251,6 +270,56 @@ describe("kernel exec - numeric instructions", () => {
       resEqual: new f32(2)
     },
 
+    {
```
maurobringolf: @xtuc First test cases are just below here. They cover `nan + value`, `nan + nan` and `value + nan`. I will add more.

maurobringolf:
> The new attributes may just be present on some nodes.

Okay thats fine too. No real value in adding a million `nan: false` and `inf: false` everywhere 👍

Issue #87, opened 2018-01-10
incorrect implementation of i32.div_s


ColinEberhardt:
The current implementation treats `div_s` and `div_u` as the same operation. However, as per the specification, `div_s` uses a 'signed' interpretation of the values:

https://webassembly.github.io/spec/core/exec/numerics.html#aux-signed

this has an impact for numbers where the most significant bit is '1'. 

To illustrate:

**unsigned**

```
(module
  (func $test (param i32) (param i32) (result i32)
   (get_local 0)
   (get_local 1)
   (i32.div_u)
  )
  (export "test" (func $test))
)
```
```
instance.exports.test(0xffffff00, 0x2);
```
```
2147483520
```

**signed**

```
(module
  (func $test (param i32) (param i32) (result i32)
   (get_local 0)
   (get_local 1)
   (i32.div_s)
  )
  (export "test" (func $test))
)
```
```
instance.exports.test(0xffffff00, 0x2);
```
```
-128
```

maurobringolf:
I was confused when I first read the code as well. I also expected the interpreter to make a distinction. Don't have time to dig in further right now though.

ColinEberhardt:
No worries. Rather than move onto reinterpret, I'll look to fix some of these bugs first.

From a general design perspective, there are a number of areas where we are going to have to apply some additional logic on top of the numeric operations that JavaScript provides. For example:

 - Handing NaN as per #46 
 - Handling i64 values
 - Handling signed / unsigned operations

We'll likely uncover more.

In the same way that we support i64 via long.js, I wonder whether it's worth adding similar types that 'boxes' all of our numeric types at runtime? i.e.

Create an interface, similar to long.js ...

```
interface Number {
  multiply(number: value): Number;
  add(number: value): Number;
  ... other binary operations
}
```

Then create x4 implementations of the above interface, one for each of the numeric types?

maurobringolf:
I like this idea. I think the main goal by using as much JS standard library and raw numbers as possible is performance? I am not sure how much impact boxing all values in objects would have on that.

From a conceptual standpoint the interface makes a lot of sense to me.

xtuc:
It's always difficult to tell with the JIT, but I assume the performance impact of "everything is an object" is insignificant compared to the gain of maintainability .

PR #90, opened 2018-01-11
[WIP] - adding a wrapper around numeric types

ColinEberhardt:
As discussed in a few recent PRs, e.g. #85, there might be some advantage to wrapping numbers are runtime. I've had a go at fixing #87 - but rather than take the easy route, have added a wrapper, then used that to apply the fix.

(please ignore the few failing tests, I wanted to get some feedback before completing this work).

The key changes are:

1. The creation of a [`Number<T>`](https://github.com/xtuc/js-webassembly-interpreter/pull/90/files#diff-fbd62d0fcf47917ebaa26c8676245ae4R78) interface. All numeric values, within `StackLocal` are implementations of this interface
2. There are currently two implementations of this interface, `BaseNumber`, which is subclassed and used by i32, f32, f64 - this uses `number` as the underlying storage. The other is `i64` which uses long.js for storage
3. This simplifies the binop code, where previously [i64 was a special case](https://github.com/xtuc/js-webassembly-interpreter/compare/master...ColinEberhardt:division-fix?expand=1#diff-68f59886a8a2af3660e543e0cc5c82ddL134).
4. Arguments passed to exported functions, and the results returned by exported functions now need wrapping / unwrapping

The div_s fix for i32 is [implemented in a BaseNumber subclass](https://github.com/xtuc/js-webassembly-interpreter/compare/master...ColinEberhardt:division-fix?expand=1#diff-240519ffcef869736d1f065ee0cf3f6fR12)

I like that this ...
 - Hides our use of long.js behind a wrapper, minimising code that is directly exposed to this external code
 - The BaseNumber subclasses could throw exceptions for unsupported operations.

For the floating point handling of NaN, this can now be implemented as a BaseNumber subclass.

I dislike that ...
 - this adds a lot of extra wrapping / unwrapping. Not a performance concern, I benchmarked before and after, and didn't see a significant change. It's more the code complexity that I worry about
  - ... although, this is mostly an impact on the tests, so perhaps not a big concern?
 - I think this makes the `type` property of StackLocal redundant? if so, should `StackLocal` still exist?

Anyhow, would be good to get some feedback on this before I do any more.

### DIFF
--- a/src/interpreter/runtime/values/i32.js
+++ b/src/interpreter/runtime/values/i32.js
@@ -1,11 +1,21 @@
 // @flow
+import { BaseNumber } from "./number";
 
 const bits = 32;
 const type = "i32";
 
-export function createValue(value: number): StackLocal {
+export class i32 extends BaseNumber {}
+
+export function createValueFromAST(value: number): StackLocal {
   value = (value | 0) % Math.pow(2, bits);
 
+  return {
+    type,
+    value: new i32(value)
+  };
+}
+
+export function createValue(value: i32): StackLocal {
   return {
     type,
     value
--- a/src/types/index.js
+++ b/src/types/index.js
@@ -75,6 +75,29 @@ type StackLocal = {
   value: any
 };
 
+interface Number<T> {
+  add(operand: T): T;
+  sub(operand: T): T;
+  mul(operand: T): T;
+  div_s(operand: T): T;
+  div_u(operand: T): T;
+  div(operand: T): T;
+  and(operand: T): T;
+  or(operand: T): T;
+  xor(operand: T): T;
+  min(operand: T): T;
+  max(operand: T): T;
+  copysign(operand: T): T;
+
+  neg(): T;
+  abs(): T;
+
+  isZero(): boolean;
+
+  equals(operand: T): boolean;
+  toNumber(): number;
+}
+
 type Label = {
   arity: number,
   value: any,
--- a/src/interpreter/index.js
+++ b/src/interpreter/index.js
@@ -1,5 +1,8 @@
 // @flow
 
+const {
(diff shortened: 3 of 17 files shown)


xtuc:
At first glance I really like this. I could look into more details tomorrow. We can then discuss about it during the video chat.

xtuc:
Do you mind having your branch on this repo instead of your fork? I wanted to rebase with my recent changes.

ColinEberhardt:
No problem at all :-)

ColinEberhardt:
Rebased onto master and completed this change - all unit tests now pass.

src/interpreter/runtime/values/i32.js
```diff
@@ -1,11 +1,21 @@
...
+import { BaseNumber } from "./number";
 
 const bits = 32;
 const type = "i32";
 
-export function createValue(value: number): StackLocal {
+export class i32 extends BaseNumber {}
+
+export function createValueFromAST(value: number): StackLocal {
   value = (value | 0) % Math.pow(2, bits);
```
xtuc: By the way, I'm not super sure about this.

The idea is to produce an in32 (`| 0` in JS or at least let's assume so :smile: ), `Math.pow(i32)` doesn't produce an i32 according to the spec.

It should probably be written:
```js
value % Math.pow(2, bits) | 0
```

src/types/index.js
```diff
@@ -75,6 +75,29 @@ type StackLocal = {
   value: any
 };
 
+interface Number<T> {
```
xtuc: I noticed that this doesn't work in Flowtype, `Number` refers to the JavaScript Number object. I will rename it `NumberInterface`.

PR #91, opened 2018-01-12
feat(instruction): implement named arguments

### LINKED ISSUE #29: Support memarg load/store instructions
The syntax is described here: https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-memarg.

WAST grammar ([spec](https://github.com/WebAssembly/spec/blob/a7e226a92e660a3d5413cfea4269824f513259d2/interpreter/README.md#s-expression-syntax)):
```bnf
offset: offset=<nat>
align: align=(1|2|4|8|...)
<type>.load((8|16|32)_<sign>)? <offset>? <align>?
<type>.store(8|16|32)? <offset>? <align>?
```

Example:

```wast
(i64.load offset=0x000 align=1 (i32.const 0))
```

xtuc:
Fixes #29

### DIFF
--- a/test/compiler/parsing/fixtures/watf/instruction/memargs/expected.json
+++ b/test/compiler/parsing/fixtures/watf/instruction/memargs/expected.json
@@ -19,8 +19,14 @@
         }
       ],
       "namedArgs": {
-        "offset": "0x000",
-        "align": "1"
+        "offset": {
+          "type": "NumberLiteral",
+          "value": 0
+        },
+        "align": {
+          "type": "NumberLiteral",
+          "value": 1
+        }
       }
     }
   ]
--- a/test/compiler/parsing/fixtures/watf/instruction/memargs/expected.json
+++ b/test/compiler/parsing/fixtures/watf/instruction/memargs/expected.json
@@ -1,0 +1,27 @@
+{
+  "type": "Program",
+  "body": [
+    {
+      "type": "Instr",
+      "id": "load",
+      "object": "i64",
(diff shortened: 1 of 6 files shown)


test/compiler/parsing/fixtures/watf/instruction/memargs/expected.json
```diff
@@ -0,0 +1,27 @@
...
+            {
+              "type": "NumberLiteral",
+              "value": 0
+            }
+          ]
+        }
+      ],
+      "namedArgs": {
+        "offset": "0x000",
+        "align": "1"
```
xtuc: So basically now we can have `key=value` pairs in instruction. This is generic enough and cover the case in #29.

Ideally I would like to have more specific types/objects for some instructions, currently the `namedArgs` bypasses the type checking.

test/compiler/parsing/fixtures/watf/instruction/memargs/expected.json
```diff
@@ -0,0 +1,33 @@
...
+      "namedArgs": {
+        "offset": {
+          "type": "NumberLiteral",
+          "value": 0
+        },
+        "align": {
+          "type": "NumberLiteral",
+          "value": 1
+        }
+      }
```
xtuc: So basically now we can have key=value pairs in instruction. This is generic enough and cover the case in #29.

Ideally I would like to have more specific types/objects for some instructions, currently the namedArgs bypasses the type checking.

PR #76, opened 2018-01-07
Feat include spec

xtuc:
- Make sure to download the gitsubmodule as well
- Usage: `make test-spec`
- Not related I implemented the WAST comments

Current status:

```
# tests 70
# pass 0
# fail 70
```

:ok_hand: (mostly because of parsing and testing helpers)

PR #100, opened 2018-01-17
feat: implements almost all of the i32 operations

### LINKED ISSUE #87: incorrect implementation of i32.div_s

ColinEberhardt:
progress towards #97 

 - [x] i32.add
 - [x] i32.sub
 - [x] i32.mul
 - [x] i32.div_u
 - [ ] i32.div_s - needs to throw 'integer overflow' exception
 - [x] i32.or
 - [x] i32.xor
 - [x] i32.and
 - [x] i32.rem_u
 - [x] i32.rem_s
 - [x] i32.shl
 - [x] i32.shr_u
 - [x] i32.shr_s
 - [x] i32.rotl
 - [x] i32.rotr
 - [x] i32.clz
 - [x] i32.ctz
 - [x] i32.popcnt
 - [x] i32.eq
 - [x] i32.eqz
 - [x] i32.ine
 - [x] i32.ilt_u
 - [x] i32.ilt_s
 - [x] i32.igt_u
 - [x] i32.igt_s
 - [x] i32.ile_u
 - [x] i32.ile_s
 - [x] i32.ige_u
 - [x] i32.ige_s

fixes #87 

I'm basically running through all of our integer operations with a goal of getting them fully spec compliant and closing off #97 

I've taken the official webassembly test suite and used a few regexes to convert it into chai / mocha test. Once #76 is implemented, these tests can be deleted.

Regarding this specific change, `i32` is now a class that implements our `NumberInterface` interface, allowing implementation of some of the specific features we need to match WebAssembly integer maths. With this PR we now support signed / unsigned division, multiplication uses Long to ensure we don't exceed JS numeric precision, and various runtime exceptions are being thrown (as per spec)


ColinEberhardt:
updated to add:
 - [x] i32.and
 - [x] i32.rem_u
 - [x] i32.rem_s

ColinEberhardt:
updated to add:
 - [x] i32.shl
 - [x] i32.shr_u
 - [x] i32.shr_s

test/interpreter/fixtures/integer-operations/exec.tjs
```diff
@@ -79,52 +79,52 @@ it("should conform to the wasm test suite", () => {
   equals(m.exports.div_u(11, 5), 2);
   equals(m.exports.div_u(17, 7), 2);
 
-  // throws(() => m.exports.rem_s(1, 0), "$4");
```
xtuc: Could we remove it instead of commenting it?

ColinEberhardt:
- [x] rotl
 - [x] rotr

src/interpreter/runtime/values/i32.js
```diff
@@ -1,14 +1,167 @@
...
 
-export class i32 extends BaseNumber {}
+// the specification describes the conversion from unsigned to signed
+// https://webassembly.github.io/spec/core/exec/numerics.html#aux-signed
+// this function performs the inverse
+const toUnsigned = a => (a < 0 ? a + Math.pow(2, bits) : a);
 
-export function createValueFromAST(value: number): StackLocal {
-  value = (value | 0) % Math.pow(2, bits);
+export class i32 implements Number<i32> {
```
xtuc: You probably need to rebase, I had to rename the interface.

xtuc:
Looks nice! I will review it in detail on my desktop tommorow.

src/interpreter/runtime/values/i32.js
ColinEberhardt: Ah - good spot, I've turned Flow back on again, and fixed the issues this PR introduced :-)

ColinEberhardt:
- [x] clz
 - [x] ctz

src/interpreter/runtime/values/number.js
```diff
@@ -69,6 +70,92 @@ export class BaseNumber implements NumberInterface<BaseNumber> {
     return new BaseNumber(-this._value);
   }
 
+  /*eslint-disable no-unused-vars*/
```
xtuc: Just FYI I commented the operand somewhere in a PR, like:
```js
rem_s(/*operand: BaseNumber*/): BaseNumber {
    throw new RuntimeError("Unsupported operation");
}
```

ColinEberhardt: Ah, good point - for some reason I thought doing that would result in Flow errors, but it does not :-)

src/interpreter/kernel/instruction/unop.js
```diff
@@ -19,6 +19,18 @@ function unop(
...
+      return createValue(c.clz());
+
+    case "ctz":
+      return createValue(c.ctz());
+
+    case "popcnt":
+      return createValue(c.popcnt());
+
+    case "eqz":
+      return createValue(c.eqz());
```
xtuc: By the way, why is the value called `c`? value sounds more explicit to me.

src/interpreter/runtime/values/i32.js
```diff
@@ -1,14 +1,271 @@
...
+        : -this._value
+    );
+  }
+
+  toNumber(): number {
+    return this._value;
+  }
+
+  isTrue(): boolean {
+    return this._value == 1;
```
xtuc: We can use triple equal here. I'm not sure about this tbh, numeric types are not booleans, either the name is misleading or we could inline the `=== 1` in the callers?

test/interpreter/fixtures/integer-operations/module.wast
```diff
@@ -0,0 +1,172 @@
...
+   (i32.ge_s)
+  )
+  (export "ge_s" (func $ge_s))
+  (func $ge_u (param i32) (param i32) (result i32)
+   (get_local 0)
+   (get_local 1)
+   (i32.ge_u)
+  )
+  (export "ge_u" (func $ge_u))
+)
```
xtuc: I really like that, the spec is way more testable this way, but way less maintainable.

src/interpreter/kernel/instruction/unop.js
ColinEberhardt: good point - I'll change that.

PR #103, opened 2018-01-19
spec test infra

xtuc:


package.json
```diff
@@ -48,7 +48,8 @@
   "bin": {
     "wasmdump": "./lib/bin/wasmdump.js",
     "wasmast": "./lib/bin/wasmast.js",
-    "wasmrun": "./lib/bin/wasmrun.js"
+    "wasmrun": "./lib/bin/wasmrun.js",
+    "wasm": "./lib/bin/repl.js"
```
xtuc: we have now a good repl

ColinEberhardt: that's an excellent idea!

ColinEberhardt:
I really like this concept, a neat solution to adding the assertions required by the spec test, with the added bonus of a REPL!

src/bin/repl.js
```diff
@@ -0,0 +1,231 @@
...
+
+  const module = instantiatedModules.find(
+    ({ exports }) => exports[name.value] !== undefined
+  );
+
+  assert(module !== undefined, `Module with export "${name.value}" not found`);
+
+  const args = exprs.map(expr => {
+    const res = partialEvaluation.evaluate(allocator, [expr]);
+    return res.value.toNumber();
```
ColinEberhardt: For i64 values `toNumber` will result in lost precision. It might be worth adding a `toString` method to the number interface and using it here.

xtuc: Yes you are right, I assume the spec tests are using i64 in function signatures (which is fine since no JS boundaries are crossed).

I will use the string representation, it's just for comparison anyway, thanks for the hint!

xtuc:
I'm going to merge this before it's too big, the spec infra is all setup now. I will be doing followup PR is order to pass the `exports.wast` test.

xtuc:
refs https://github.com/xtuc/js-webassembly-interpreter/issues/98

Issue #63, opened 2018-01-06
Implement `reinterpret` instructions


xtuc:
- [ ] `i32.reinterpret`
- [ ] `i64.reinterpret`
- [ ] `f32.reinterpret`
- [ ] `f64.reinterpret`

See specification here: https://webassembly.github.io/spec/core/exec/numerics.html#op-reinterpret

ColinEberhardt:
I'll take this - I'd like to use it as a way to test #79

PR #105, opened 2018-01-25
implemented f32 / f64 reinterpret

ColinEberhardt:
I've made some progress towards #63 - this PR implements `i32.reinterpret/f32`, again using a transformed version of the WebAssembly spec tests.

The most notable change here is that I've split the `NumberInterface` interface into three separate interfaces, reflecting the different operations that are available on floats vs. integers.

**UPDATE**

I've completed `i32.reinterpret/f32` and `i64.reinterpret/f64` - with the f64 / i64 functionality tested via unit tests rather than the spec level tests

### DIFF
--- a/src/interpreter/kernel/instruction/unop.js
+++ b/src/interpreter/kernel/instruction/unop.js
@@ -1,6 +1,14 @@
 // @flow
 
-type Sign = "abs" | "neg" | "clz" | "ctz" | "popcnt" | "eqz";
+type Sign =
+  | "abs"
+  | "neg"
+  | "clz"
+  | "ctz"
+  | "popcnt"
+  | "eqz"
+  | "reinterpret/f32"
+  | "reinterpret/f64";
 
 import * as i32 from "../../runtime/values/i32";
 import * as i64 from "../../runtime/values/i64";
@@ -31,6 +39,12 @@ function unop(
 
     case "eqz":
       return createValue(value.eqz());
+
+    case "reinterpret/f32":
+      return i32.createValue(value.reinterpret());
+
+    case "reinterpret/f64":
+      return i64.createValue(value.reinterpret());
   }
 
   throw new Error("Unsupported unop: " + sign);
--- a/src/interpreter/kernel/instruction/unop.js
+++ b/src/interpreter/kernel/instruction/unop.js
@@ -1,6 +1,6 @@
 // @flow
 
-type Sign =
+type Operation =
   | "abs"
   | "neg"
   | "clz"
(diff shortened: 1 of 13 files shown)


src/types/AST.js
```diff
@@ -3,7 +3,7 @@
 import type { i32 } from "../interpreter/runtime/values/i32";
 
 type U32Literal = NumberLiteral & {
-  value: NumberInterface<i32>
+  value: NumericOperations<i32>
```
ColinEberhardt: This doesn't look right to me, this interface defines operations that can be performed on a numeric type. I don't think that is appropriate for the AST, which I see as just a data representation.

src/interpreter/kernel/instruction/unop.js
```diff
@@ -31,6 +38,9 @@ function unop(
 
     case "eqz":
       return createValue(value.eqz());
+
+    case "reinterpret/f32":
+      return i32.createValue(value.reinterpret());
```
ColinEberhardt: this breaks the current pattern, which expects all operations to return the same type as their operand(s). Also, this shouldn't be hard-coded as `i32`

src/interpreter/kernel/instruction/unop.js
```diff
@@ -1,6 +1,13 @@
 // @flow
 
-type Sign = "abs" | "neg" | "clz" | "ctz" | "popcnt" | "eqz";
+type Sign =
```
xtuc: It seems that the naming is wrong here

src/interpreter/kernel/instruction/unop.js
xtuc: Yeah interesting, it returns an opaque `StackLocal` tho. I think there's nothing conceptually wrong, but just confusing.

src/interpreter/runtime/values/number.js
```diff
@@ -1,167 +1,87 @@
...
   _value: number;
 
   constructor(value: number) {
     this._value = value;
   }
 
-  add(operand: BaseNumber): BaseNumber {
-    return new BaseNumber(this._value + operand._value);
+  add(operand: Float<U>): Float<U> {
+    return new this.constructor(this._value + operand._value);
```
xtuc: That an interesting hack here :smile:

src/types/AST.js
xtuc: This is not clear enough, could we discuss about that on Monday next week?

ColinEberhardt: yeah, no worries

src/interpreter/kernel/instruction/unop.js
ColinEberhardt: Yeah, I think it bends the current patten a little, but not too much.

src/interpreter/kernel/instruction/unop.js
ColinEberhardt: Yeah, good point, I've renamed to `Operation`

src/interpreter/runtime/values/number.js
ColinEberhardt: Yeah, polymorphism / inheritance isn't JavaScript's strong point!

ColinEberhardt:
Sure, no problem

PR #106, opened 2018-01-25
Export global instance

xtuc:


src/interpreter/runtime/values/module.js
```diff
@@ -125,6 +136,27 @@ export function createInstance(
           createModuleExportIntance(node.name, externalVal)
         );
       }
+
+      if (node.descr.type === "Global") {
+        const instantiatedGlobal = instantiatedGlobals[node.descr.id.value];
+
+        if (instantiatedGlobal.type.mutability === "var") {
+          throw new CompileError("Mutable globals cannot be exported");
```
xtuc: This is an early validation error.

src/interpreter/runtime/values/module.js
```diff
@@ -125,6 +136,27 @@ export function createInstance(
...
+        }
+
+        if (instantiatedGlobal.type.valtype === "i64") {
+          throw new LinkError("Export of globals of type i64 is not allowed");
```
xtuc: Chrome also throws a LinkError

PR #122, opened 2018-01-29
Reinterpret f32 nan

maurobringolf:
I think I will add `inf` cases in a separate PR. I believe the current logic does not support `-nan` yet, but I have not verified that yet.

Also I converted the parser to use a `signature` which I will extract to the types module where we can define signature for instructions as arrays, i.e. `[ 'i32', 'i32' ]` for `i32.add`. Let me know what you think about that @xtuc. I need to check where `object` is used and make sure that will still be correctly supported.


src/interpreter/runtime/values/f32.js
```diff
@@ -42,6 +42,30 @@ export class f32 extends Float<i32> {
 }
 
 export class f32nan extends f32 {
+  reinterpret(): i32 {
```
ColinEberhardt: it would be useful if you could add a comment which references the relevant section in the spec that defines this behaviour - thanks!

maurobringolf: That's a good point, but I could not find much in the spec about it. The conversion directly reflects the  IEEE 754-2008 standard for `nan`s, but I linked it anyways. Lmk if you think this is helpful or more confusing.

maurobringolf:
I am trying to run the spec tests for nan conversions @ColinEberhardt converted into this suite here:

https://github.com/xtuc/js-webassembly-interpreter/blob/reinterpret-nan/test/interpreter/fixtures/conversions/exec.tjs#L431-L442

However, these don't work yet. I figured out that they call `f32.reinterpret` instead of `f32nan.reinterpret` but I have no idea why. Any thoughts?

maurobringolf:
**UPDATE:** I figured out why these tests are not working. They use my hex parse function from `webassembly-floating-point-hex-parser` which does support `nan` and `inf` (the floating point hex format does not).

src/interpreter/runtime/values/f32.js
```diff
@@ -42,6 +42,39 @@ export class f32 extends Float<i32> {
...
+
+    // sign bit of _value shifted to position 0
+    if (this._value <= 0) {
+      result = result | 0x80000000;
+    }
+
+    // 8-bit exponent shifted to position 1 through 8
+    result = result | (0xff << 23);
+
+    // 23-bit mantissa which is obtained by disregaring the sign of _value
```
xtuc: disregaring?

maurobringolf: *disregarding, good catch

maurobringolf:
@ColinEberhardt What do you think about deleting the ones moved into kernel/exec from the fixtures file?

ColinEberhardt:
> What do you think about deleting the ones moved into kernel/exec from the fixtures file?

I'm all for removing unit tests that basically verify the same thing, which is why I'm happy to see the test I created, based on the 'official' spec, deleted, now that we can run the spec test directly:

https://github.com/xtuc/js-webassembly-interpreter/pull/116/files#diff-3706c05f16b3e9dd04e96d9e40083617

The spec tests provide a decent overall coverage, which means our unit tests can be selective and targeted.

Issue #59, opened 2018-01-06
Implement memory operation instructions


xtuc:
Load:
- [x] `i32.load`
- [x] `i64.load`
- [x] `f32.load`
- [x] `f64.load`
- [x] `i32.load8_s`
- [x] `i32.load8_u`
- [x] `i32.load16_s`
- [x] `i32.load16_u`
- [x] `i64.load8_s`
- [x] `i64.load8_u`
- [x] `i64.load16_s`
- [x] `i64.load16_u`
- [x] `i64.load32_s`
- [x] `i64.load32_u`

Store:
- [x] `i32.store`
- [x] `i64.store`
- [x] `f32.store`
- [x] `f64.store`
- [x] `i32.store8`
- [x] `i32.store16`
- [x] `i64.store8`
- [x] `i64.store16`
- [x] `i64.store32`

See specification here https://webassembly.github.io/spec/core/exec/instructions.html#exec-load.

This depends on https://github.com/xtuc/js-webassembly-interpreter/issues/29

ColinEberhardt:
I have a question about how this might be implemented. With the current code you can supply a `Memory` instance via the module imports:

https://github.com/xtuc/js-webassembly-interpreter/blob/master/src/interpreter/index.js#L59

However, this seems to be used as the 'allocator' which the `ModuleInstance` uses at runtime to store things like functions and globals.

My understanding is that WebAssembly linear memory is separate from the runtime stack, so I wouldn't expect runtime related information to appear in the `Memory` I supply to my WebAssembly  module.

Have I correctly understood the current implementation?

xtuc:
Now that i'm thinking about it there might be a misunderstand on my side.

Currently, you can either provide an WebAssembly.Memory (at instantiation) or a default one will be created, they are both used as backend of the allocator and available at runtime for memory allocations.

xtuc:
:tada:

PR #123, opened 2018-01-30
Implemented i32.store and i64.store

ColinEberhardt:
A complete implementation of  i32.store and i64.store #59 

- The allocator no longer uses the runtime memory, it instead uses an array which represents the store (as per the spec)
- Memory now uses an `ArrayBuffer`
- 
- The `store` instruction accesses this via `frame.originatingModule.memaddrs[0]`, as defined in the spec: https://webassembly.github.io/spec/core/exec/instructions.html#exec-storen

You can see it fully working here:

https://github.com/xtuc/js-webassembly-interpreter/pull/123/files#diff-6d5a2351c718a8fe03cfcfa79ca3fd90

### DIFF
--- a/src/compiler/parsing/watf/grammar.js
+++ b/src/compiler/parsing/watf/grammar.js
@@ -1,4 +1,5 @@
 // @flow
+import { parse32I } from "./number-literals";
 
 const { tokens, keywords } = require("./tokenizer");
 const t = require("../../AST");
@@ -159,11 +160,11 @@ export function parse(tokensList: Array<Object>, source: string): Program {
         );
       }
 
-      const limits = t.limits(token.value);
+      const limits = t.limits(parse32I(token.value));
       eatToken();
 
       if (token.type === tokens.number) {
-        limits.max = token.value;
+        limits.max = parse32I(token.value);
         eatToken();
       }
 
--- a/src/interpreter/kernel/exec.js
+++ b/src/interpreter/kernel/exec.js
@@ -1,9 +1,10 @@
+import { Memory } from "../runtime/values/memory";
+
 // @flow
 
 const {
   binopi32,
   binopi64,
-
   binopf32,
   binopf64
 } = require("./instruction/binop");
@@ -111,10 +112,10 @@ export function executeStackFrame(
       );
     }
 
-    if (c1.type !== type2) {
+    if (c1.type !== type1) {
       throw new RuntimeError(
         "Internal failure: expected c1 value of type " +
-          type2 +
+          type1 +
           " on top of the stack, give type: " +
           c1.type
       );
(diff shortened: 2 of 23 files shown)


src/interpreter/kernel/exec.js
```diff
@@ -1,3 +1,5 @@
+import { debuglog } from "util";
```
xtuc: Ah yes good point. We could also use it here https://github.com/xtuc/js-webassembly-interpreter/blob/3dde1533bec38620ef95931dfa66955b9b798beb/src/interpreter/index.js#L199-L205 to trace the execution.

src/interpreter/kernel/memory.js
```diff
@@ -6,38 +6,41 @@ export const NULL = 0x0;
 // FIXME(sven): remove this ^
 export const ptrsize = 1;
```
xtuc: I don't think this is actually used, I wanted to fake the behavior of low-level pointer but that's maybe unnecessary?

src/interpreter/kernel/exec.js
```diff
@@ -703,6 +705,21 @@ export function executeStackFrame(
         return pop1();
       }
 
+      /**
+       * Memory operations
+       */
+      case "store": {
+        const memAddr = frame.originatingModule.memaddrs[0];
```
xtuc: Is the memory of the module always at index 0?

src/interpreter/kernel/exec.js
```diff
@@ -703,6 +705,21 @@ export function executeStackFrame(
...
+        const memory = (frame.allocator.get(memAddr): Memory);
```
xtuc: In Flow this is actually not a cast, you want:

```js
const memory: Memory = frame.allocator.get(memAddr);
```

src/interpreter/kernel/exec.js
ColinEberhardt: yes, it is documented in the `store` instruction here: https://webassembly.github.io/spec/core/exec/instructions.html#exec-storen

and the design docs too: http://webassembly.org/docs/modules/#linear-memory-section

src/interpreter/kernel/exec.js
ColinEberhardt: thanks - need to read up a bit more on Flow :-)

src/interpreter/kernel/memory.js
ColinEberhardt: That's an interesting point - I guess with the real runtime the store (which the allocator uses) is a proper heap, sorting the funcs, globals etc ... whereas the allocator in this emulator stores references, rather that the data structures themselves. 

I do like the way the emulator closely mimics the real runtime.

src/interpreter/kernel/exec.js
xtuc: Right, thanks for the clarification.

src/interpreter/runtime/values/module.js
```diff
@@ -92,8 +92,26 @@ export function createInstance(
     },
 
     Memory({ node }: NodePath<Memory>) {
-      // TODO(sven): implement exporting a Memory instance
-      const memoryinstance = null;
+      const limits = {
```
ColinEberhardt: @xtuc it looks like limits are represented as strings within the AST - is that right? should they be `number`? also they should be able to handle hex representation

xtuc: If that's the case it's a bug, they should be number (or hex).

They are numbers in the AST
https://github.com/xtuc/js-webassembly-interpreter/blob/master/src/types/AST.js#L203-L204

xtuc: You are right! https://github.com/xtuc/js-webassembly-interpreter/blob/64851bba4ca86b1fa9e277099d0c3ffe13de6bfc/test/compiler/parsing/fixtures/watf/table/basic/expected.json#L21-L22

It's a bug.

src/compiler/parsing/watf/grammar.js
```diff
@@ -159,11 +160,11 @@ export function parse(tokensList: Array<Object>, source: string): Program {
         );
       }
 
-      const limits = t.limits(token.value);
+      const limits = t.limits(parse32I(token.value));
```
ColinEberhardt: @xtuc - I've added a fix for the 'limits' bug. Not sure whether I should be importing numeric parsing logic into `grammar.js` - or whether `t.limits` should take string args and perform this parsing logic. Any thoughts?

xtuc: Conceptually t.* shouldn't have too much logic, they're just there to construct the AST.

xtuc: Conceptually t.* shouldn't have too much logic, they're just there to construct the AST.

xtuc: Conceptually t.* shouldn't have too much logic, they're just there to construct the AST.

PR #127, opened 2018-01-31
Reinterpret inf in f32 and f64

maurobringolf:
This adds the necessary code and tests for the following four cases:

```
(assert_return (invoke "i64.reinterpret_f64" (f64.const -inf)) (i64.const 0xfff0000000000000))
(assert_return (invoke "i64.reinterpret_f64" (f64.const inf)) (i64.const 0x7ff0000000000000))
(assert_return (invoke "i32.reinterpret_f32" (f32.const inf)) (i32.const 0x7f800000))
(assert_return (invoke "i32.reinterpret_f32" (f32.const -inf)) (i32.const 0xff800000))
```

The code (also the one for `nan` reinterpretation) contains some operations on constants like `0xff <<23` which I left in there for the comments to make some sense and general readability. I believe these are optimized anyway either by the minifier or the engine (constants folding) anyway, but let me know if you prefer to only have the pre-computed results in there.

*Sorry for the `make fix` commits with every PR* 😁

### DIFF
--- a/src/interpreter/runtime/castIntoStackLocalOfType.js
+++ b/src/interpreter/runtime/castIntoStackLocalOfType.js
@@ -19,12 +19,12 @@ export function castIntoStackLocalOfType(
     f64: f64.createValueFromAST
   };
 
-  if (nan) {
+  if (nan === true) {
     castFn.f32 = f32.createNanFromAST;
     castFn.f64 = f64.createNanFromAST;
   }
 
-  if (inf) {
+  if (inf === true) {
     castFn.f32 = f32.createInfFromAST;
     castFn.f64 = f64.createInfFromAST;
   }
--- a/src/interpreter/runtime/castIntoStackLocalOfType.js
+++ b/src/interpreter/runtime/castIntoStackLocalOfType.js
@@ -15,8 +15,10 @@ export function castIntoStackLocalOfType(
   const castFn = {
     i32: i32.createValueFromAST,
     i64: i64.createValueFromAST,
-    f32: inf ? f32.createInfFromAST : (nan ? f32.createNanFromAST : f32.createValueFromAST),
-    f64: inf ? f64.createInfFromAST :f64.createValueFromAST
+    f32: inf
+      ? f32.createInfFromAST
+      : nan ? f32.createNanFromAST : f32.createValueFromAST,
+    f64: inf ? f64.createInfFromAST : f64.createValueFromAST
   };
 
   if (typeof castFn[type] === "undefined") {
--- a/src/interpreter/runtime/castIntoStackLocalOfType.js
+++ b/src/interpreter/runtime/castIntoStackLocalOfType.js
@@ -15,11 +15,19 @@ export function castIntoStackLocalOfType(
   const castFn = {
     i32: i32.createValueFromAST,
     i64: i64.createValueFromAST,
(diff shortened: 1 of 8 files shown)


src/interpreter/runtime/castIntoStackLocalOfType.js
```diff
@@ -9,13 +9,16 @@ const f64 = require("./values/f64");
...
+  inf: boolean = false
 ): StackLocal {
   const castFn = {
     i32: i32.createValueFromAST,
     i64: i64.createValueFromAST,
-    f32: nan ? f32.createNanFromAST : f32.createValueFromAST,
-    f64: f64.createValueFromAST
+    f32: inf
+      ? f32.createInfFromAST
+      : nan ? f32.createNanFromAST : f32.createValueFromAST,
```
xtuc: I'm not a super fan of terner in terner :(

maurobringolf: Good point, I thought maybe we could pass the whole AST node into `createValueFromAST`, so I dont need to pass `inf` and `nan` flags around and everything is handled inside one method. What do you think?

xtuc: Yeah, like `AST Node To StackLocal` :+1:

maurobringolf: I replaced the ternaries with if statements for now, changing the signature of `createValueFromAST` will be a bigger refactor which I'd like to do in a separate pr.

src/interpreter/runtime/castIntoStackLocalOfType.js
```diff
@@ -9,15 +9,26 @@ const f64 = require("./values/f64");
...
 ): StackLocal {
   const castFn = {
     i32: i32.createValueFromAST,
     i64: i64.createValueFromAST,
-    f32: nan ? f32.createNanFromAST : f32.createValueFromAST,
+    f32: f32.createValueFromAST,
     f64: f64.createValueFromAST
   };
 
+  if (nan) {
```
xtuc: Could you please check for `nan === true and same for inf, it's a perf gain

maurobringolf: Ah cool! Any references on that? Ill do it before merging

xtuc: Yes it's bascially the ToBoolean algorythm which is called here to coerce `nan` to a boolean, if you do `nan === true` it doesn't need to.

maurobringolf: Makes sense, thanks!

src/interpreter/runtime/castIntoStackLocalOfType.js
```diff
@@ -9,15 +9,26 @@ const f64 = require("./values/f64");
...
+    castFn.f32 = f32.createNanFromAST;
+    castFn.f64 = f64.createNanFromAST;
```
xtuc: I noticed that `f64.createNanFromAST` doesn't exist.

maurobringolf: I added the placeholders, but there is no `nan` code for `f64` yet.

PR #131, opened 2018-02-03
implementation of all load operations 🎉 

ColinEberhardt:
The code within `exec.js` is a bit switch / case heavy, and I was itching to refactor - but felt this was probably a decent implementation from a performance perspective.

This branch is a continuation of #130 - I'd suggest reviewing  merging that first.


src/compiler/parsing/watf/number-literals.js
```diff
@@ -54,7 +54,7 @@ export function parse32I(sourceString: string): number {
 export function parse64I(sourceString: string): LongNumber {
   let long: Long;
   if (isHexLiteral(sourceString)) {
-    long = Long.fromString(sourceString.substr(2), false, 16);
+    long = Long.fromString(sourceString, false, 16);
```
ColinEberhardt: the use of `substr` was removing the sign, and wasn't needed - Long can cope with negative hex values

src/interpreter/kernel/exec.js
```diff
@@ -747,6 +772,83 @@ export function executeStackFrame(
...
+      case "load":
+      case "load16_s":
+      case "load16_u":
+      case "load8_s":
+      case "load8_u":
+      case "load32_s":
+      case "load32_u": {
+        const memory = getMemory();
+
+        const ptr = pop1("i32").value.toNumber() + getMemoryOffset(instruction);
```
xtuc: Just an idea but the `pop{1,2}` is performing a validation from the spec, should we rename to `validateAndPop{1,2}` to make it clear? (don't need to do this in this pr).

src/interpreter/kernel/exec.js
```diff
@@ -170,6 +172,32 @@ export function executeStackFrame(
...
+        throw new RuntimeError("offset must be positive");
+      }
+      if (offset > 0xffffffff) {
+        throw new RuntimeError(
+          "offset must be less than or equal to 0xffffffff"
+        );
+      }
+      return offset;
+    } else {
+      return 0;
```
xtuc: IRC I had a constant `NULL = 0x0`, we can use it for every memory operations if it can make things clearer?

xtuc:
Have you tried with the spec tests btw?

xtuc:
I wanted to help you here but this branch is on your fork :(

src/interpreter/kernel/exec.js
ColinEberhardt: I think most of the helper methods within `exec` will perform some validation, so I wouldn't bother renaming them :-)

ColinEberhardt:
> Have you tried with the spec tests btw? 

Yes, and they didn't work! they relied on other features, data sections, imports etc ...

I think I'll make the memory spec tests my next priority

src/interpreter/runtime/values/f32.js
```diff
@@ -101,3 +101,13 @@ export function createValue(value: f32): StackLocal {
     value
   };
 }
+
+export function createValueFromArrayBuffer(
```
ColinEberhardt: I'm not that keen on this pattern, but would like to get this merged, and refactor later - perhaps using a similar pattern to binop / unop

PR #134, opened 2018-02-06
use Exception to signal a trap

xtuc:


test/interpreter/kernel/exec/administrative-instructions.js
```diff
@@ -67,29 +50,12 @@ describe("kernel exec - administrative instructions", () => {
     const stackFrame = createStackFrame(code, []);
     stackFrame.trace = (_, x) => (pc = x);
 
-    executeStackFrame(stackFrame);
-
-    assert.notEqual(code.length, pc + 1);
-    assert.equal(pc, 1);
-  });
-
-  it("should stop executing the stackframe at unreachable in child and propagate up the stack", () => {
```
xtuc: It's mentioned in the spec that a trap "bubble up the stack" but it's seems not exposed to the caller.

I was worried about loose our virtual execution stack but we can collect it and attach it to the TrapException Object.

PR #136, opened 2018-02-07
Implement reinterpret for f64 nans

maurobringolf:
This covers the following cases:

```wasm
(assert_return (invoke "i64.reinterpret_f64" (f64.const -nan:0xfffffffffffff)) (i64.const -1))
(assert_return (invoke "i64.reinterpret_f64" (f64.const nan)) (i64.const 0x7ff8000000000000))
(assert_return (invoke "i64.reinterpret_f64" (f64.const -nan)) (i64.const 0xfff8000000000000))
(assert_return (invoke "i64.reinterpret_f64" (f64.const nan:0x4000000000000)) (i64.const 0x7ff4000000000000))
(assert_return (invoke "i64.reinterpret_f64" (f64.const -nan:0x4000000000000)) (i64.const 0xfff4000000000000))
```

### DIFF
--- a/src/interpreter/runtime/values/f64.js
+++ b/src/interpreter/runtime/values/f64.js
@@ -37,11 +37,31 @@ export class f64inf extends f64 {
       upper = upper | 0x80000000;
     }
 
-    return new i64(Long.fromBits(0, upper));
+    return new i64(Long.fromBits(0, upper).toSigned());
   }
 }
 
-export class f64nan extends f64 {}
+export class f64nan extends f64 {
+  reinterpret(): i64 {
+    let lower = 0;
+    let upper = 0;
+
+    // sign bit of _value shifted to position 0
+    if (this._value <= 0) {
+      upper = upper | 0x80000000;
+    }
+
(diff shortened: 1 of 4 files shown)


src/interpreter/runtime/values/f64.js
```diff
@@ -37,11 +37,31 @@ export class f64inf extends f64 {
...
+      upper = upper | 0x80000000;
+    }
+
+    // 11-bit exponent shifted to position 1 through 11
+    upper = upper | (0x7ff << 20);
+
+    // 52-bit mantissa which is obtained by disregarding the sign of _value
+    const mantissa = this._value <= 0 ? -this._value : this._value;
+    lower = lower | (mantissa % (2 ** 32))
+    upper = upper | Math.floor(mantissa / (2 ** 32))
```
maurobringolf: Note that I couldn't use bitshift like in the `f32` case here because the payload is 52 bits and we only have bitwise for 32 bits. These are the mathematical equivalents as far as I know, but I only tested with the provided tests.

PR #139, opened 2018-02-08
First try at WAST printer

xtuc:
## TODO

- Make it lint (Flow don't get it)
  - I think we should remove the flowtype check

---

@johnsudaar do you think we need an unified interface for our printers?

The wasm printer can't use the same recursive code gen tho since it's divided into sections.


xtuc:
Looks good to me, it's a first implementation we can iterate on this later.

src/compiler/printer/wast/index.js
```diff
@@ -0,0 +1,442 @@
+// @flow
+/* eslint flowtype-errors/show-errors: warn */
```
xtuc: I had to disable flow unfortunately, something is not working with the AST nodes definition.

PR #142, opened 2018-02-11
feat: share page

xtuc:


xtuc:
It's live! :tada: 

We can now share binaries: [add.wasm](https://webassembly.js.org/share.html#0,97,115,109,1,0,0,0,1,7,1,96,2,127,127,1,127,3,2,1,0,7,7,1,3,97,100,100,0,0,10,9,1,7,0,32,0,32,1,106,11)

Issue #147, opened 2018-02-15
globals should not support multiple initialisation instructions


ColinEberhardt:
I can't find this documented in the official specification, however the 'unofficial' specification indicates that the `init` field of globals (and data sections), is an 'instantiation-time initializer'

https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#global-section

These are single instructions, therefore the way we handle globals is currently incorrect, as illustrated by this test:

https://github.com/xtuc/js-webassembly-interpreter/blob/master/test/compiler/parsing/fixtures/watf/global/with-multiple-instrs-in-init/actual.wast

I've verified this by using the wat2wasm tool on the acbove `actual.wast` file, and it fails with the message:

```
error: invalid global initializer expression, must be a constant expression; either *.const or get_global.
  (global i32
```

xtuc:
I added a note here about that https://github.com/xtuc/js-webassembly-interpreter/blob/431caff0cb1debecd7c3b864103f78d3debc1ac6/src/interpreter/runtime/values/global.js#L19-L23. Might be worth opening an issue on the spec to clarify this?

The test you mentioned is only for parsing, since it's an 'instantiation-time initializer' it should still parse, right? My implementation of global initialization passes the spec tests tho.

ColinEberhardt:
raised an issue with the spec here: https://github.com/WebAssembly/spec/issues/694

xtuc:
Thanks!

PR #150, opened 2018-02-15
feat: support parsing of data sections

ColinEberhardt:
This is my work-in-progress for supporting data segment at the parser / AST level.

Work remaining:
- [x] support the optional memory index (currently I've hard-coded to zero)

As I've not done any work on the parser / AST before it would be good to get some feedback, especially on the AST representation I've gone for.


src/compiler/AST/index.js
```diff
@@ -393,7 +393,7 @@ export function memory(limits: Limit, id: ?Index): Memory {
 
 export function data(
   memoryIndex: Index,
-  offset: Array<Instruction>,
+  offset: Instruction,
```
ColinEberhardt: see #147 - data segments also only support a single instruction

src/compiler/parsing/watf/grammar.js
```diff
@@ -193,6 +194,48 @@ export function parse(tokensList: Array<Object>, source: string): Program {
...
+     *
+     * WAST:
+     *
+     * data:  ( data <index>? <offset> <string> )
+     */
+    function parseData(): Data {
+      eatTokenOfType(tokens.openParen);
+
+      let offset: Instruction;
+      if (token.type === tokens.valtype) {
```
ColinEberhardt: the logic I've added to parse a `const` or `get_global` feels like it must be repeating some code elsewhere - but I just can't find the right code to re-use!

test/compiler/parsing/index.js
```diff
@@ -21,6 +21,7 @@ function createCheck(suite, ast) {
   let expected;
   try {
     expected = readFileSync(expectedFile, "utf8");
+    expected = JSON.stringify(JSON.parse(expected), null, 2);
```
ColinEberhardt: prettier was formatting my expected JSON files, causing the tests to fail. This change re-formats the expected JSON to match the actual.

src/compiler/parsing/watf/grammar.js
xtuc: Yeah that's right, I would prefer to implement the parser (+ tests) before trying to re-use code.

test/compiler/parsing/index.js
xtuc: We can probably ignore them in the prettier configuration? I guess you are using the integration in your editor because I don't have that issue?

src/compiler/parsing/watf/grammar.js
ColinEberhardt: cool, makes sense.

test/compiler/parsing/index.js
ColinEberhardt: yes, it is due to prettier - however, I think it makes the tests a bit less fragile if we don't have to formal the expected JSON in a specific way.

ColinEberhardt:
Thanks  for the feedback, I'll keep progressing this ...

On another note, Flow is now providing useful feedback, the build error indicates that I have broken the WASM parser due to a change in AST format. 👍

test/compiler/parsing/index.js
xtuc: Ok, yes that's right.

src/compiler/AST/index.js
xtuc: I didn't it for WAST where the definition is: `data:    ( data <var>? ( offset <instr>* ) <string>* )`.

I'm using some transform (https://webassembly.js.org/docs/contrib-transform-wast-identifier-to-index.html) to handle these differences.

src/compiler/parsing/watf/string-literals.js
```diff
@@ -0,0 +1,82 @@
...
+    case "′":
+      return 0x27;
+    case "\\":
+      return 0x5c;
+  }
+  return -1;
+}
+
+const ESCAPE_CHAR = 92; // backslash
+const QUOTE_CHAR = 34; // backslash
```
xtuc: Just FYI we use https://github.com/xtuc/charcodes/tree/master/packages/charcodes in Babel's parser. We can use it also here if it make it more clearer.

ColinEberhardt:
This PR should hopefully be ready for merge - I've implemented the logic which writes the data segment to the module's linear memory.

Currently it only supports constant expressions for the offset, I don't think we have support for global imports yet.

xtuc:
> I don't think we have support for global imports yet.

We do, see https://github.com/xtuc/js-webassembly-interpreter/blob/master/test/interpreter/fixtures/import-const-global/module.wast.

src/interpreter/runtime/values/module.js
```diff
@@ -437,6 +466,8 @@ export function createInstance(
 
   instantiateInternals(n, allocator, instantiatedInternals, moduleInstance);
 
+  instantiateDataSections(n, allocator, moduleInstance);
```
xtuc: `instantiateDataSections` should be in internal IMO, because it's instantiating the module's internal. It's probably better so because it will hurt the readability, but also adds a additional traversal. What do you think?

xtuc:
Also the wasm decoder can be another PR :+1:

src/interpreter/runtime/values/module.js
ColinEberhardt: The reason I added this additional traversal is because [the spec mentions that](https://webassembly.github.io/spec/core/text/modules.html#text-module):

> A module consists of a sequence of fields that can occur in any order.

Therefore, a data section can appear before the memory or global it refers to.

The logic could be moved to `instantiateInternals`, but would require some form of post-traversal processing.

ColinEberhardt:
>> I don't think we have support for global imports yet.

> We do, see https://github.com/xtuc/js-webassembly-interpreter/blob/master/test/interpreter/fixtures/import-const-global/module.wast.

Thanks for the heads-up, that helps. Although, it looks like the following syntax isn't supported:

```
(global $a (import "m" "a") i32)
```

xtuc:
Yes that's correct, i'm working on this.

src/interpreter/runtime/values/module.js
xtuc: Ah that's perfect then! thanks for the clarification

xtuc:
It it see https://github.com/xtuc/js-webassembly-interpreter/blob/master/test/compiler/parsing/fixtures/watf/global/with-shorthand-import/actual.wast and your example parsed.

Maybe your branch is not up-to-date with master?

PR #154, opened 2018-02-19
Revert "fix: disable IfInstruction tests"

xtuc:
This reverts commit cb8757395ca5fc740288af2eb34e1556068da0f3.

Continuation of #140

PR #158, opened 2018-02-20
br instruction part 2 (and type instruction parsing)

xtuc:
We need `br` to run the spec tests for #154


xtuc:
Currently fails:

```
  143 |
  144 |   (type $sig (func (param i32 i32 i32) (result i32)))
> 145 |   (table anyfunc (elem $f))
      |                     ^
  146 |   (func (export "as-call_indirect-func") (result i32)
  147 |     (block (result i32)
  148 |       (call_indirect (type $sig)
Error: Assertion error: expected token of type closeParen, given keyword
```

xtuc:
Currently the evaluation of the instructions is using a tree, that's not convenient we jumping to another code location because it will keep its parent.

src/interpreter/kernel/exec.js
```diff
@@ -15,11 +15,13 @@ const i64 = require("../runtime/values/i64");
 const f32 = require("../runtime/values/f32");
 const f64 = require("../runtime/values/f64");
 const label = require("../runtime/values/label");
-const { createChildStackFrame } = require("./stackframe");
+const stackframe = require("./stackframe");
 const { createTrap } = require("./signals");
 const { RuntimeError } = require("../../errors");
 const t = require("../../compiler/AST");
 
+// DEFINE_MACRO(CHECK_STACK_UNWIND, 'if (frame._unwindReason !== 0) { break; }');
```
xtuc: I'm trying to get a macro system working, I'm looking for a c-like macro system.
The existing one in Babel doesn't work since Babel can't parse the illegal `break` statement here "lol".

xtuc:
Don't review it for now, I made a super ugly implementation of the stack unwinding mecasim.

PR #164, opened 2018-02-21
Start section/func

xtuc:
refs https://github.com/webpack/webpack/pull/6531#issuecomment-367380500


xtuc:
Left todo is adding the spectest, I created an issue https://github.com/xtuc/js-webassembly-interpreter/issues/165 to track it

PR #167, opened 2018-02-23
Fix various stuff from #158 

xtuc:
Cutting out some features from #158


src/compiler/parsing/wasm/decoder.js
```diff
@@ -31,6 +31,17 @@ const {
...
+/**
+ * FIXME(sven): we can't do that because number > 2**53 will fail here
+ * because they cannot be represented in js.
+ */
+function badI32ToI64Conversion(value: number): LongNumber {
+  return {
+    high: value < 0 ? -1 : 0,
+    low: value >>> 0
+  };
+}
```
xtuc: I'll create an issue about that

src/compiler/parsing/wasm/decoder.js
```diff
@@ -743,7 +755,12 @@ export function decode(ab: ArrayBuffer, printDump: boolean = false): Program {
 
           dump([value], "value");
 
-          args.push(t.numberLiteral(value));
+          const node = {
+            type: "LongNumberLiteral",
+            value: badI32ToI64Conversion(value)
+          };
+
+          args.push(node);
```
xtuc: We're missing an AST node builder for LongNumberLiteral.

src/interpreter/kernel/exec.js
```diff
@@ -921,15 +968,22 @@ export function executeStackFrame(
...
   if (frame.values.length > 0) {
-    return pop1();
+    const res = pop1();
+
+    if (res.type !== "label") {
+      return res;
+    } else {
+      // Push label back
+      pushResult(res);
+    }
```
xtuc: That's a problem with the current working of the interpreter, we should end up with a label here.

PR #186, opened 2018-02-28
wasm-edit

xtuc:
### Notes
- Optimization: remove empty section after a removal
- When adding an empty section, ensure the correct section order


packages/wasm-gen/src/encoder/index.js
```diff
@@ -0,0 +1,69 @@
...
+  bytes[0] = v;
+
+  return bytes;
+}
+
+export function encodeHeader(): Array<Byte> {
+  return constants.magicModuleHeader;
+}
+
+// FIXME(sven): implement unsigned LEB128 encoder here
```
xtuc: @xtuc todo here

PR #198, opened 2018-03-12
doc: minor documentation tweaks

ColinEberhardt:
Just familiarising myself with the project again, post-lerna. This is a very small change that helps 'sign post' the various sub-projects.


xtuc:
Thanks for your pr and yes sorry, I couldn't add much doc (mostly because I'm spending my time in the Webpack integration).

We can also setup a call if that helps.

PR #205, opened 2018-03-18
Exponential floats

maurobringolf:
A first approximation towards implementing float exponential literals ( `1e1`, etc). Very rough and lots of things uncovered, but I am working on a larger refactor and better test coverage of the tokenizer.

I also setup direct tokenizer tests by basically copy pasting the ast assertion code. This could probably be unified into one implementation that works for both (Maybe pass transformation function (`parse`, `tokenize`) and test paths (`fixtures`, `tokenize`) to a function and reuse the logic.


packages/wast-parser/src/tokenizer.js
```diff
@@ -266,7 +266,9 @@ function tokenize(input: string) {
 
       while (
         numberLiterals.test(char) ||
-        (input[current - 1] === "p" && char === "+")
+        (input[current - 1] === "p" && char === "+") ||
```
ColinEberhardt: Nice job! As a general observation, would it be possible to unit test this code more directly? Testing it via wat and expected AST produces a lot of additional 'noise'

maurobringolf: Yes, this is what the new tokenizer tests are for. I aim to split the tests for string->tokens and tokens->ast as much as possible.

xtuc: Yes, that sounds good to me.

packages/wast-parser/test/fixtures/number-literals/f32/negative-exponential/expected.json
```diff
@@ -0,0 +1,27 @@
...
+      "params": [],
+      "result": [],
+      "body": [
+        {
+          "type": "Instr",
+          "id": "const",
+          "object": "f32",
+          "args": [
+            {
+              "type": "FloatLiteral",
```
xtuc: I noticed something when I was working on the `wast-printer`, we don't preserve the original notation. For example: `0x01` becomes `1` because we store it as a number in the AST. I would propose introducing a `raw` fields that captures the raw string, we can then use this when printing the AST.

What do you think? We can do it in another PR since it impact some other places.

packages/wast-parser/src/tokenizer.js
maurobringolf: There are so many cases for number literals :roll_eyes: I will look into generating an exhaustive test set from the spec tests.

packages/wast-parser/test/fixtures/number-literals/f32/negative-exponential/expected.json
maurobringolf: Yep it makes tests more readable and definitely a good idea for the printer :+1: I open an issue for it.

PR #206, opened 2018-03-18
Implement lookhead() and lookback() as functions [->lookbehind()]

maurobringolf:
This introduces two helper functions `lookahead()` and `lookback()` to look at input characters without consuming them. It builds on #205 

What do you think of this direction in general? There is a lot of fiddling with the `current` pointer and direct references to `input` that it is sometimes hard to tell whats going on. I'd like to convert most of these operations into small utility functions in order to make the "main" code more readable.

@xtuc Did you avoid helper functions on purpose or is this something that makes sense to you? In C I would make these helper functions macros to make sure there is no overhead, which would technically be possible here too with `babel-macros`. But I'd rather avoid that if possible.

### DIFF
--- a/packages/wast-parser/src/tokenizer.js
+++ b/packages/wast-parser/src/tokenizer.js
@@ -115,13 +115,13 @@ function tokenize(input: string) {
   /**
    * Can be used to look at the next character(s).
    *
-   * The default behavior `peek()` simply returns the next character without consuming it.
+   * The default behavior `lookahead()` simply returns the next character without consuming it.
    *
    * @param int length How many characters to query. Default = 1
    * @param int offset How many characters to skip from current one. Default = 1
    *
    */
-  function peek(length = 1, offset = 1) {
+  function lookahead(length = 1, offset = 1) {
     return input.substring(current + offset, current + offset + length);
   }
 
@@ -134,7 +134,7 @@ function tokenize(input: string) {
     let char = input[current];
 
     // ;;
-    if (char === ";" && peek() === ";") {
+    if (char === ";" && lookahead() === ";") {
       eatToken();
       eatToken();
 
@@ -160,7 +160,7 @@ function tokenize(input: string) {
     }
 
     // (;
-    if (char === "(" && peek() === ";") {
+    if (char === "(" && lookahead() === ";") {
       eatToken(); // (
       eatToken(); // ;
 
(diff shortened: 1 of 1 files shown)


xtuc:
> @xtuc Did you avoid helper functions on purpose or is this something that makes sense to you? In C I would make these helper functions macros to make sure there is no overhead, which would technically be possible here too with babel-macros. But I'd rather avoid that if possible.

I preferred duplicating some code in the parser (mostly because at the time wast<>wat was not clear to me), but I definitely agree that we should have helper functions.

On a side note, I searched for C-like marco in JS for the kernel/exec file, and noticed something interesting. Babylon (the Babel parser) and probably most of the JS parsers have the `unsyntactic continue/break` error, which prevents you from emending a `return`, `continue` or `break` in your replacement. So, we rather should use helper function.

packages/wast-parser/src/tokenizer.js
```diff
@@ -112,6 +112,32 @@ function tokenize(input: string) {
...
+  /**
+   * Can be used to look at the last few character(s).
+   *
+   * The default behavior `lookback()` simply returns the last character.
+   *
+   * @param int length How many characters to query. Default = 1
+   * @param int offset How many characters to skip back from current one. Default = 1
+   *
+   */
+  function lookback(length = 1, offset = 1) {
```
xtuc: Could we rename it to `lookbehind`? It sounds more common to me.

xtuc:
Note that I also switch the base of the branch against your first PR #205 for clarity. Feel free to change it once the #205 has been merged.

packages/wast-parser/src/tokenizer.js
maurobringolf: Yeah that sounds better :+1:

maurobringolf:
> On a side note, I searched for C-like marco in JS for the kernel/exec file, and noticed something interesting. Babylon (the Babel parser) and probably most of the JS parsers have the unsyntactic continue/break error, which prevents you from emending a return, continue or break in your replacement. So, we rather should use helper function.

I am not sure if I understand this correctly, do you have any links/references?

xtuc:
Sorry that wasn't clear, my issue was when I declared my macro as code, where Babylon was used to parse the code and failed because of the unexpected `break`. You can fix this by passing a string but I can remember why this didn't work out well either.

PR #207, opened 2018-03-18
Additional validation from the wat parser

ColinEberhardt:
while running `wat2wasm` over various test cases I've spotted some rules that are not being applied. Currently there is validation logic within the interpreter, but not the wat-parser. As these rules seems to be enforced by `wat2wasm` I thought it was best to enforce them in the parser.

What do you think?

I've got a few other tests to fix as well, but thought it was worth getting some feedback.

### DIFF
--- a/packages/webassemblyjs/test/index.js
+++ b/packages/webassemblyjs/test/index.js
@@ -15,7 +15,7 @@ function toArrayBuffer(buf) {
 describe("interpreter", () => {
   describe("wasm", () => {
     const testSuites = glob.sync(
-      "packages/WebAssembly/test/fixtures/**/module.wasm"
+      "packages/webassemblyjs/test/fixtures/**/module.wasm"
     );
 
     testSuites.forEach(suite => {
@@ -47,7 +47,7 @@ describe("interpreter", () => {
 
   describe("wat", () => {
     const testSuites = glob.sync(
-      "packages/WebAssembly/test/fixtures/**/module.wast"
+      "packages/webassemblyjs/test/fixtures/**/module.wast"
     );
 
     testSuites.forEach(suite => {
--- a/packages/ast/src/traverse.js
+++ b/packages/ast/src/traverse.js
@@ -71,8 +71,13 @@ function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
     case "Data":
     case "Memory":
     case "Elem":
+    case "FuncImportDescr":
+    case "GlobalType":
     case "NumberLiteral":
+    case "ValtypeLiteral":
     case "FloatLiteral":
+    case "StringLiteral":
+    case "QuoteModule":
     case "LongNumberLiteral":
     case "BinaryModule":
     case "LeadingComment":
@@ -127,7 +132,10 @@ function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
     case "ModuleImport": {
       cb(n.type, createPath(n, parentPath));
 
-      cb(n.descr.type, createPath(n.descr, parentPath));
+      if (n.descr != null) {
+        // $FlowIgnore
+        walk(n.descr, cb, createPath(n, parentPath));
+      }
 
       break;
     }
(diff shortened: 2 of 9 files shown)


maurobringolf:
Awesome, this looks great to me. Are these specific to wat or could these be placed in one of the other packages to be reused?

ColinEberhardt:
> Awesome, this looks great to me. Are these specific to wat or could these be placed in one of the other packages to be reused?

Good question, however, the structure of wasm, with clearly defined sections, might require a different approach to validation.

> Related: I have been reading a bit on validation of types and if I understand correctly it should be possible to implement a pass that verifies that all instruction stack argument and return value type match up. Do we have anything for this setup?

That's a very good point, we certainly have some wat files in the interpreter tests that fail this check ;-)

I'm planning to add additional validation as I go along - the main thing to debate initially is whether this belongs in wat-parser.

xtuc:
> Related: I have been reading a bit on validation of types and if I understand correctly it should be possible to implement a pass that verifies that all instruction stack argument and return value type match up. Do we have anything for this setup?

Yes, I started it here https://github.com/xtuc/webassemblyjs/blob/master/packages/webassemblyjs/src/compiler/validation/type-inference.js.

packages/wast-parser/src/validator.js
```diff
@@ -0,0 +1,47 @@
...
+  let startFuncs = 0;
+
+  traverse(ast, {
+    Start() {
+      startFuncs++;
+    }
+  });
+
+  if (startFuncs > 1) {
+    throw new Error("wat modules can only contain one Start function");
```
xtuc: For my own debugging purposes, I started using the `babel-code-frame` package which allows us to print the location in the code along with the error. In the future we could re-implement it with our own printer or with more information in the parser.

ColinEberhardt:
Thanks for the feedback @xtuc & @maurobringolf what's the best route forward for this PR?

> Looks great! I'm wondering if it would make sense to have a new package @webassemblyjs/validations that provide these validations?

Should I migrate these validations, and those from the interpreter, before merging?

xtuc:
> Should I migrate these validations, and those from the interpreter, before merging?

We an do it after your merge, I can also create the package if you want.

ColinEberhardt:
>> Should I migrate these validations, and those from the interpreter, before merging?

> We an do it after your merge, I can also create the package if you want.

OK, cool - I'll fix up the tests and merge. Thanks

ColinEberhardt:
@xtuc I've fixed up the tests, but this uncovered a few more issues which i have resolved. You might want to review again before I merge.

The changes are:
- AST traversal additional recursion to create a proper path for module import nested statements
- with global shorthand imports the parser previously moved the import outside of the global within the AST. I can't see any way the relationship between the two was preserved? I've modified this behaviour so that the import is nested within the global in the `init` property. 

I also think the `init` property of a global should not be an array, as per the rules discussed here: https://github.com/WebAssembly/spec/issues/694 - but I've not made that change yet.

packages/wast-parser/src/index.js
```diff
@@ -1,13 +1,14 @@
...
+import validate from "./validator";
 const { tokenize } = require("./tokenizer");
 
 export function parse(source: string): Program {
   const tokens = tokenize(source);
 
   // We pass the source here to show code frames
   const ast = parser.parse(tokens, source);
-
+  validate(ast);
```
xtuc: Actually, I don't know what's the best place for this check. I used to run the validations here but it annoyed me when I wanted to just test the parser. I moved it right in the module "compilation".

What do you think? It's not a parse error but a validation one.

ColinEberhardt: Yeah, that’s a good point, having validation in multiple places will confused things. I initially added it here to mirror the functionality of wat2wasm,but that doesn’t mean it’s right for this project. I’ll move it :-)

xtuc: ok thanks

packages/webassemblyjs/test/index.js
```diff
@@ -15,7 +15,7 @@ function toArrayBuffer(buf) {
 describe("interpreter", () => {
   describe("wasm", () => {
     const testSuites = glob.sync(
-      "packages/WebAssembly/test/fixtures/**/module.wasm"
+      "packages/webassemblyjs/test/fixtures/**/module.wasm"
```
ColinEberhardt: @xtuc I was wondering why the total number of unit tests had dropped recently ;-)

xtuc: omg, that's a good catch!

PR #208, opened 2018-03-18
[WIP] Implement and use eatCharacter() method

maurobringolf:
This refines the helper function `eatToken` by implementing `eatCharacter` instead. Couple of thoughts about the current implementation which led to this PR:

* We are reading characters and produce a list of tokens. Technically the `eatToken` method eats characters, so I find its current name confusing.

* The method `eatToken` does not update the `char` value, which leads to the two patterns `eatToken(); char = input[current]` and `char = input[current++]` which are both breaking abstraction (which makes correcting `column` values harder for example).

My solution is a function `nextCharacter(amount = 1)` which updates `current`, `column` and `char` values. I was able to replace almost all occurences of the two patterns mentioned above without any issue. There are still a few cases left which is why this is a WIP. So:

- [x] Replace all `eatToken` calls and eventually remove the function itself.

- [x] Wait for #206 and resolve conflicts.

- [x] If every possible branch either calls `eatCharacter` or throws an error we can remove this line too https://github.com/xtuc/webassemblyjs/pull/208/files#diff-e3c84c771a65bfbdb0559ea1726f2699R127.


packages/wast-parser/src/tokenizer.js
```diff
@@ -117,21 +118,23 @@ function tokenize(input: string) {
     current++;
   }
 
+  function eatCharacter(amount = 1) {
+    column += amount;
+    char = input[(current += amount)];
```
xtuc: Could you please split this into two lines?

packages/wast-parser/src/tokenizer.js
```diff
@@ -113,6 +96,30 @@ function tokenize(input: string) {
...
   const tokens = [];
 
+  /**
+   * Creates a pushToken function for a given type
+   */
+  function pushToken(type: string) {
+    return function(v: string | number, opts: Object = {}) {
+      tokens.push(Token(type, v, line, column, opts));
+    };
+  }
```
maurobringolf: @xtuc What do you think of this? This way `column` and `line` values are automatically inserted via closure.

maurobringolf:
I am not sure why this is failing, the errors are due to old testfiles related to https://github.com/xtuc/webassemblyjs/pull/210.

I would like to merge this as is and iterate. The column will still be wrong in most cases and the logic is a bit scattered, but now we can work on it by adding individual tokenizer tests.

packages/wast-parser/src/tokenizer.js
xtuc: Yup :+1:

xtuc:
I fixed the tests on master, feel free to merge it. :+1:

maurobringolf:
Awesome, thanks!

PR #212, opened 2018-03-20
Decode custom name section

### LINKED ISSUE #202: wasm parser generates unique ids, wat parser does not
Based on my work in #201, I'm starting to look at the differences in AST generated by the wasm and wat parser / decoders. 

Taking the following file:

https://github.com/xtuc/webassemblyjs/blob/master/packages/webassemblyjs/test/fixtures/fabs/module.wast

Here is the diff between the AST generated by (1) wast-parser, (2) wat2wasm => wasm-parser:

```diff
 - Expected
+ Received

  {
    "type": "Program",
    "body": [
      {
        "type": "Module",
        "id": null,
        "fields": [
          {
            "type": "Func",
            "name": {
              "type": "Identifier",
-             "value": "abs"
+             "value": "func_1"
            },
            "params": [
              {
                "valtype": "f32"
              }
            ],
            "result": [
              "f32"
            ],
            "body": [
              {
                "type": "Instr",
                "id": "get_local",
                "args": [
                  {
                    "type": "NumberLiteral",
                    "value": 0
                  }
                ]
              },
              {
                "type": "Instr",
                "id": "abs",
                "object": "f32",
                "args": []
              }
            ]
          },
          {
            "type": "ModuleExport",
            "name": "abs",
            "descr": {
              "type": "Func",
              "id": {
                "type": "Identifier",
-               "value": "abs"
+               "value": "func_1"
              }
            }
          }
        ]
      }
    ]
 }
```

The difference is that the wasm decoder re-writes function identifiers with unique IDs:

https://github.com/xtuc/webassemblyjs/blob/master/packages/wasm-parser/src/decoder.js#L463

Why does the wasm parser have this logic, but not the wat parser? 

i.e. should I ignore this difference in the tests, or fix it!

ColinEberhardt:
This PR adds logic to the wasm-parser to decode custom name sections, fixes #202 

Changes in this PR:
 - wasm-parser tests no longer have binary wasm files as test cases, they instead use wat files which are transformed to wasm via `wat2wasm` (I'm not sure how we make this a proper dependency?)
 - wasm-parser parses the custom name section generated by `wat2wasm --debug-names`. Currently it only parses the function name sub-section. I'll look at adding local names next. Names are stored in the module-metadata (which I've restructured a bit to make it easier to walk).
 - wasm-parser walks the AST, after decoding, to apply names. I've added an option to disable this behaviour as desired, as there will be a performance impact.

TODO:
 - [x] work out how make `wat2wasm` available to travis
 - [x] fix flow errors!

### DIFF
--- a/packages/ast/src/index.js
+++ b/packages/ast/src/index.js
@@ -33,6 +33,32 @@ export function signature(object: string, name: string): SignatureMap {
   return sign[0];
 }
 
+export function functionNameMetadata(
+  value: string,
+  index: number
+): FunctionNameMetadata {
+  return {
+    type: "FunctionNameMetadata",
+    value,
+    index
+  };
+}
+
+export function moduleMetadata(
+  sections: Array<SectionMetadata>,
+  functionNames: Array<FunctionNameMetadata>
+): ModuleMetadata {
+  const n: ModuleMetadata = {
+    type: "ModuleMetadata",
+    sections
+  };
+
+  if (functionNames.length) {
+    n.functionNames = functionNames;
+  }
+  return n;
+}
+
 export function identifier(value: string): Identifier {
   return {
     type: "Identifier",
--- a/packages/wasm-parser/src/decoder.js
+++ b/packages/wasm-parser/src/decoder.js
@@ -62,12 +62,15 @@ function byteArrayEq(l: Array<Byte>, r: Array<Byte>): boolean {
 export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
   const buf = new Uint8Array(ab);
 
-  let inc = 0;
+  const inc = {};
 
   function getUniqueName(prefix: string = "temp"): string {
-    inc++;
-
-    return prefix + "_" + inc;
+    if (!inc.hasOwnProperty(prefix)) {
+      inc[prefix] = 0;
+    } else {
+      inc[prefix] = inc[prefix] + 1;
+    }
+    return prefix + "_" + inc[prefix];
   }
 
   let offset = 0;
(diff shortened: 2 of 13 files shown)


packages/ast/src/index.js
```diff
@@ -33,6 +33,32 @@ export function signature(object: string, name: string): SignatureMap {
...
+  const n: ModuleMetadata = {
+    type: "ModuleMetadata",
+    sections
+  };
+
+  if (functionNames.length) {
+    n.functionNames = functionNames;
+  }
+  return n;
+}
```
xtuc: I'm not sure the metadata should be a node in our AST? I already model it here https://github.com/xtuc/webassemblyjs/blob/7b3e82af4bf89b49af46cac5d64e8659a06b8037/packages/ast/src/index.js#L93 as a property on the module, what do you think?

packages/wasm-parser/src/decoder.js
```diff
@@ -1388,17 +1447,23 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
...
+        const sectionName = readUTF8String();
+        eatBytes(sectionName.nextIndex);
 
-        return { nodes, metadata };
+        const remainingBytes = sectionSizeInBytes - sectionName.nextIndex;
+
+        if (sectionName.value === "name") {
+          metadata.push(...parseNameSection(remainingBytes));
+        } else {
+          // We don't parse otehr custom section
```
xtuc: nit: typo `otehr` -> `other`

packages/wasm-parser/src/decoder.js
```diff
@@ -959,6 +962,59 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
...
+    const subSectionSizeInBytes = subSectionSizeInBytesu32.value;
+    eatBytes(subSectionSizeInBytesu32.nextIndex);
+
+    // TODO: add support for naming locals - for now we just eat this section
+    eatBytes(subSectionSizeInBytes);
+  }
+
+  // this is a custom suction that wat2wasm includes if invoked
+  // using the --debug-names option
+  function parseNameSection(remainingBytes: number) {
```
xtuc: I'm wondering if we shouldn't split it in a `wabt` file or something? Because it's specific to the binary they're emitting.

Also how we could know that it has been emitted by `wabt`?

packages/wasm-parser/src/index.js
```diff
@@ -1,14 +1,51 @@
...
 
 const defaultDecoderOpts = {
   dump: false,
   ignoreCodeSection: false,
-  ignoreDataSection: false
+  ignoreDataSection: false,
+  ignoreCustomNameSection: false
 };
 
+function restoreNames(ast) {
```
xtuc: Instead of doing a separate pass couldn't we do this in the decoder directly?

xtuc:
>  work out how make wat2wasm available to travis

We could use https://www.npmjs.com/package/wast2wasm and wrap it into an executable.

packages/ast/src/index.js
ColinEberhardt: I've kept the metadata in the same place. The change I've made is to give it a `type` property so that it can be traversed by the AST parser. Does that sound OK?

packages/wasm-parser/src/decoder.js
ColinEberhardt: good point - my next step will be to complete this work by supporting locals, so I could have a go at that then?

packages/wasm-parser/src/index.js
ColinEberhardt: I don't think this is possible. The name section appears at the very end of the module, so the only way to apply the naming is via a second pass.

packages/wasm-parser/src/decoder.js
xtuc: Yes sure

ColinEberhardt:
@xtuc I've updated and fixed the tests - it now uses wabt.js rather than the executable. There are a few unresolved discussion points, would be interested in your feedback.

After this, I'll look at decoding local names.

Then, the final goal, run the interpreter test fixtures,  against the wasm-parser!

packages/wasm-parser/src/index.js
xtuc: Ok that make sense.

According to the spec a custom section can be after each non-custom section, so depending on how the binary was constructed It could happens, but we can leave as is since it will work in any case.

packages/wasm-parser/src/decoder.js
```diff
@@ -959,6 +962,59 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
...
+    const functionNames = [];
+
+    const subSectionSizeInBytesu32 = readU32();
+    eatBytes(subSectionSizeInBytesu32.nextIndex);
+
+    const numbeOfFunctionsu32 = readU32();
+    const numbeOfFunctions = numbeOfFunctionsu32.value;
+    eatBytes(numbeOfFunctionsu32.nextIndex);
+
+    for (let i = 0; i < numbeOfFunctions; i++) {
```
xtuc: Here's a small typo: `numbeOfFunctions` -> `numberOfFunctions`

ColinEberhardt:
Ooops - I made a mistake in the files which I was round-tripping. Need to make a few more changes. Although now that the tests are correct it is starting to throw up differences in the wat and wasm parsers:

https://github.com/xtuc/webassemblyjs/issues/218

ColinEberhardt:
I've simplified this PR to just include the custom name section logic - I'll add the wasm-parser test work to a different PR as this is starting to look a like a bigger job

Issue #156, opened 2018-02-20
Use testsuite repo instead of the entire spec


xtuc:
The spec tests are available here https://github.com/WebAssembly/testsuite, the repo is smaller.

PR #214, opened 2018-03-21
Number literals: Generate and pass tests

### LINKED ISSUE #213: Preserve '_' in number literals in rawValue of ast node
Currently the tokenizer deletes all numeric separators _ and they are therefore lost in the asts rawValue attribute. This can be fixed by moving the deletion from the tokenizer into the parser.

maurobringolf:
I extracted as many integers as I could from [int_literals.wast](https://raw.githubusercontent.com/WebAssembly/spec/master/test/core/int_literals.wast) and generated `actual.wast` and `expected.json` test files for the tokenizer from them. For clarity I also checked in the scripts that do that (`scripts/download-number-literals.sh` and `scripts/generate-tokenizer-tests.js`) so things can be reproduced or tweaked as needed.

This is only integer literals and the only observable change I had to make to pass them all is fixing the issue with numeric separators in the AST. So this fixes #213.


packages/wast-parser/src/tokenizer.js
```diff
@@ -126,26 +126,32 @@ function tokenize(input: string) {
...
    *
    * @param int length How many characters to query. Default = 1
    * @param int offset How many characters to skip forward from current one. Default = 1
    *
    */
   function lookahead(length = 1, offset = 1) {
-    return input.substring(current + offset, current + offset + length);
+    return input
+      .substring(current + offset, current + offset + length)
+      .toLowerCase();
```
xtuc: I'm wondering if this could slow down the lookup process?

packages/wast-parser/src/tokenizer.js
```diff
@@ -293,26 +300,23 @@ function tokenize(input: string) {
...
-      if (char === "0" && lookahead().toUpperCase() === "X") {
+      if (char === "0" && lookahead() === "x") {
         value += "0x";
         numberLiterals = HEX_NUMBERS;
         eatCharacter(2);
       }
 
       while (
-        numberLiterals.test(char) ||
+        (char !== undefined && numberLiterals.test(char)) ||
```
xtuc: I think this case only happens in the tests, like: `0x0`, but in real you wouldn't encounter that.

What do you think if we wrap your test into `()`?

packages/wast-parser/test/tokenizer/number-literals/+42/expected.json
```diff
@@ -0,0 +1 @@
+[{"type":"number","value":"+42","loc":{"start":{"line":1,"column":1}}}]
```
xtuc: To improve the readability, could we please use `JSON.stringify(x, null, 2)` here?

scripts/download-number-literals.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# integer literals
+curl https://raw.githubusercontent.com/WebAssembly/spec/master/test/core/int_literals.wast \
+	| head -n70 \
+	| grep -oP "(i(32|64).const\K\s*((\d|-|x|_|\+)*))" \
+	> packages/wast-parser/test/tokenizer/raw/int_literals.txt
+
```
xtuc: We could fetch it from the submodules in this repo.

packages/wast-parser/test/tokenizer/number-literals/+42/expected.json
maurobringolf: Oh I didn't know that it supports formatting :sunglasses: Thanks!

scripts/download-number-literals.sh
maurobringolf: I'm not familiar with Git submodules, how would you do that?

packages/wast-parser/src/tokenizer.js
maurobringolf: I thought about that, but bad input can then result in an infinite loop which should be avoided imo

packages/wast-parser/src/tokenizer.js
maurobringolf: Probably. The reason I moved it here is because every comparison of a lookahead with a single character (`lookahead() === `X`) had to deal with it separately which is much more error prone (easy to forget).

xtuc: Ok, I get that. We can still change it later.

scripts/download-number-literals.sh
xtuc: We can leave as is for now, because there's #156 and it's better to have that decoupled.

packages/wast-parser/src/tokenizer.js
```diff
@@ -293,26 +300,37 @@ function tokenize(input: string) {
...
-        (lookbehind().toUpperCase() === "E" && char === "-") ||
-        (value.length > 0 && char.toUpperCase() === "E")
+        (lookbehind() === "p" && char === "-") ||
+        (lookbehind() === "e" && char === "+") ||
+        (lookbehind() === "e" && char === "-") ||
+        (value.length > 0 && (char === "e" || char === "E"))
       ) {
         if (char === "p" && value.includes("p")) {
-          throw new Error("Unexpected character `p`.");
+          throw new Error('Unexpected character "p"');
```
xtuc: We can maybe have a magic helper for this, like `throwUnexpectedChar` which would take the char with a closure.

PR #215, opened 2018-03-24
Rewrite number tokenizer as separate state machine

maurobringolf:
Okay so this is an experiment on top of #214 . This other PR passes a lot of tests because it allows **way too many strings** as number literals. It is largely based on regex, one while loop and then if statements during each iteration. I want to make it more precise such that malformed literals actually produce errors. Once I started going down the [the list in the spec tests](https://github.com/WebAssembly/spec/blob/master/test/core/float_literals.wast#L195) I realized that this is incredibly difficult. Almost each case has to be handled separately.

So I decided to do an experiment and rewrite the number tokenizer as state machine. I find it easier to reason about, because we can look at transitions from state to state individually. The code is not structured properly (move things out of the loop, better names etc), but I want to check if you're open to something like this? @xtuc


packages/wast-parser/src/tokenizer.js
```diff
@@ -282,56 +290,127 @@ function tokenize(input: string) {
...
+
+      const NUMERIC_SEPARATOR = "_";
+
+      let state = START;
+      let eatLength = 1;
+
+      const dbg = () => unexpectedCharacter(char);
+      const regexToState = (r, s, n = 1, allowSeparators = false) => () => {
+        if (allowSeparators) {
+          debugger;
```
xtuc: aha, you probably forgot that one

maurobringolf:
@xtuc @ColinEberhardt Apart from lint and flow errors it passes all tests. I tried to clean it up but I am still not super happy with the result. What do you think? Do you have any experience in writing fsm's?

xtuc:
> Do you have any experience in writing fsm's?
nope

ColinEberhardt:
> Do you have any experience in writing fsm's

Nope - me neither! However, I've had a quick look at the code, and it makes a lot of sense to me.  For a given state you provide an array of regexes, each of which result in a given state transition. Nice and simple!

packages/wast-parser/src/tokenizer.js
```diff
@@ -165,6 +175,144 @@ function tokenize(input: string) {
...
+      match = transitions[i]();
+      if (match) {
+        break;
+      }
+    }
+
+    return match;
+  };
+
+  const START = "START";
```
ColinEberhardt: should this be a union of string literals, i.e.

```
type States = 'START' | 'HEX' | ...
```

packages/wast-parser/src/tokenizer.js
```diff
@@ -165,6 +175,144 @@ function tokenize(input: string) {
...
+  };
+
+  /**
+   * Combines a list of transition functions into a complete state transition function.
+   * When none of the given transitions apply, the given default state is returned.
+   *
+   * @param {Array} transitions A list of transition functions
+   * @param {String} defaultState The state to be returned in case no transition matches
+   *
+   * @return {Array} A two element array containing [newState, eatLength] where eatLength is the amount of characters to consume.
```
ColinEberhardt: This is a flow tuple type https://flow.org/en/docs/types/tuples/

packages/wast-parser/src/tokenizer.js
```diff
@@ -165,6 +175,144 @@ function tokenize(input: string) {
...
+    if (regex.test(lookahead(n, 0))) {
+      return [nextState, n];
+    }
+
+    return false;
+  };
+
+  /**
+   * Combines a list of transition functions into a complete state transition function.
+   * When none of the given transitions apply, the given default state is returned.
```
ColinEberhardt: Is it worth adding that the order of transitions matters? i.e. the first that matches is the 'winner'

packages/wast-parser/src/tokenizer.js
```diff
@@ -165,6 +175,144 @@ function tokenize(input: string) {
...
+    ),
+    DEC_SIGNED_EXP: combineTransitions(
+      [regexToState(/\+|-/, DEC_EXP), regexToState(/[0-9]/, DEC_EXP)],
+      STOP
+    ),
+    DEC_EXP: combineTransitions([regexToState(/[0-9]/, DEC_EXP, 1, DEC_EXP)]),
+
+    HEX: combineTransitions(
+      [
+        regexToState(/[0-9|A-F|a-f]/, HEX, 1, HEX),
```
ColinEberhardt: isn't the fourth arg of regexToState a boolean?

packages/wast-parser/src/tokenizer.js
maurobringolf: Definitely, that is important. Thanks Ill add it

packages/wast-parser/src/tokenizer.js
maurobringolf: Ah forgot to update the docs there: The problem is that regexToState needs to stay in the current state when a separator is encountered, but it doesnt have access to the state variable here. So I pass the state instead of the boolean, but Id prefer this to be a boolean. Ill see if I can do that

packages/wast-parser/src/tokenizer.js
maurobringolf: Thanks for the flow hints, I have to go over it and do the types thats right.

maurobringolf:
Update: I added flow typing and am now trying to extract the FSM logic into a separate file which will hopefully make the tokenizer more understandable and the FSM potentially reusable.

maurobringolf:
@xtuc I managed to extract the finite state machine logic into a separate file. It implements a generic kind of finite state machine that we could possibly reuse for other text parsing things (identifiers and strings for example).

The only concern I have is that the remaining `input` is copied (passed to the FSM via `input.slice()`) for every number literal parse step. But this could be avoided by wrapping the input in an object (maybe an `Input` class that implements `lookahead` and `eat` in a future PR?). Anyway, it'd be really cool if we can merge this.

packages/wast-parser/test/tokenizer/number-literal-errors/multiple-colons/actual.wast
```diff
@@ -1 +0,0 @@
-nan::1234
```
xtuc: Are these tests not possible anymore?

packages/wast-parser/src/fsm.js
```diff
@@ -0,0 +1,98 @@
...
+
+function makeTransition<Q>(
+  regex: RegExp,
+  nextState: Q,
+  n: number = 1,
+  allowedSeparator: ?string
+): TransitionEdge<Q> {
+  return function() {
+    if (allowedSeparator) {
+      if (this.input[this.ptr] === allowedSeparator) {
```
xtuc: The `this` looks a bit mystic to me. Could you rather pass the instance as argument in the function?

packages/wast-parser/src/fsm.js
```diff
@@ -0,0 +1,98 @@
...
+type State<T> = T | Symbol;
+type TransitionEdge<Q> = () => [Q, number] | false;
+type TransitionFunction<Q> = () => [Q, number];
+type TransitionList<Q> = { [Q]: Array<TransitionEdge<Q>> };
+
+function makeTransition<Q>(
+  regex: RegExp,
+  nextState: Q,
+  n: number = 1,
+  allowedSeparator: ?string
```
xtuc: The API would be clearer to me if you passed an object here for `n` and `allowedSeparator`.

Its usage would then be:
```js
makeTransition(/[0-9]/, "DEC_FRAC", { n: 1, allowedSeparator: NUMERIC_SEPARATOR }),
```

packages/wast-parser/test/fsm/index.js
```diff
@@ -0,0 +1,23 @@
...
+  it("simple fsm", () => {
+    const simpleFsm = new FSM(
+      {
+        A: [makeTransition(/a/, "B")],
+        B: [makeTransition(/b/, "A")]
+      },
+      "A"
+    );
+    assert.strictEqual(simpleFsm.run("hello"), "");
+    assert.strictEqual(simpleFsm.run("ahello"), "a");
```
xtuc: We can also add a test for `abhello ⟶ ab`

packages/wast-parser/src/tokenizer.js
```diff
@@ -279,62 +331,14 @@ function tokenize(input: string) {
       char === "-" ||
       char === "+"
     ) {
-      let value = "";
-      const startColumn = column;
+      const value = numberLiteralFSM.run(input.slice(current));
```
xtuc: Above we test if it's a literal before going through the FSM. Is there a way to handle errors into the FSM so we can remove the pre-condition here?

packages/wast-parser/src/tokenizer.js
```diff
@@ -86,8 +86,69 @@ const keywords = {
...
+    DEC: [
+      makeTransition(/[0-9]/, "DEC", 1, NUMERIC_SEPARATOR),
+      makeTransition(/\./, "DEC_FRAC"),
+      makeTransition(/e|E/, "DEC_SIGNED_EXP")
+    ],
+    DEC_SIGNED_EXP: [
+      makeTransition(/\+|-/, "DEC_EXP"),
+      makeTransition(/[0-9]/, "DEC_EXP")
+    ],
+
```
xtuc: nit: extra nl here

xtuc:
Also i'm wondering if we couldn't parse the `NumberToken` in the tokenizer? We could include it in the FSM? It seems a bit redundant because we have the same kind of logic to actually parse them later.

packages/wast-parser/src/fsm.js
maurobringolf: Very good point, this way I don't need to pass `n: 1` for every separator, thanks!

packages/wast-parser/src/tokenizer.js
maurobringolf: Hmm I have to think about this, not sure. We could run the fsm at this point and check whether the output is an empty string or not. I don't think that will behave completely the same currently, but it should be doable.

I prefer this solution, because I think running the fsm is expensive (currently copying complete remaining string) so it should only be ran when we know theres a number literal. I'm not completely sure about the perf aspect though

packages/wast-parser/src/fsm.js
maurobringolf: Sadly not as easy, because the consuming code provides transitions as arguments to construct an `FSM` instance, so theres a bootstrapping problem. I solved it by binding all transition functions in `combineTransitions`, but its not the prettiest agreed. Couldnt think of a better solution yet.

packages/wast-parser/test/tokenizer/number-literal-errors/multiple-colons/actual.wast
maurobringolf: Ahh not sure why these are removed, I'll adjust and add them back in. Sorry

packages/wast-parser/src/fsm.js
```diff
@@ -10,7 +10,8 @@ type TransitionList<Q> = { [Q]: Array<TransitionEdge<Q>> };
 function makeTransition<Q>(
   regex: RegExp,
   nextState: Q,
-  { n = 1, allowedSeparator }: { n: number, allowedSeparator: ?string } = {}
+  // $FlowIgnore
+  { n = 1, allowedSeparator } = {}
```
maurobringolf: I couldnt get flow to accept this. I tried 
```
{ n = 1, allowedSeparator } : { n: number, allowedSeparator: ?string } = {}
```
without success. (All calls to the function are marked with "incompatible type with object literal").

packages/wast-parser/src/fsm.js
maurobringolf: Sorry, I probably misunderstood. Your solution is indeed possible and I used it  9afce4c8a976a5ab4f23be519e4ac8a9a7da5fdc

packages/wast-parser/test/tokenizer/number-literal-errors/multiple-colons/actual.wast
maurobringolf: I restored them 7095053.

maurobringolf:
> Also i'm wondering if we couldn't parse the NumberToken in the tokenizer? We could include it in the FSM? It seems a bit redundant because we have the same kind of logic to actually parse them later

Yep we are loosing some information currently. We could try and use the FSM to add an attribute indicating the type of numberToken (int, decimal, hex, hex fractional, exponential, hex floating point etc) or even make them different token types. Then this code in `number-literals.js` could be replaced by simpler less redundant checks for the attribute or token type:

https://github.com/xtuc/webassemblyjs/blob/master/packages/wast-parser/src/number-literals.js#L79-L101

But parsing shouldnt be done entirely in the tokenize function imo.

packages/wast-parser/src/fsm.js
xtuc: I think the correct type is `{n: number, allowedSperator?: string }` since allowedSperator is optional.

packages/wast-parser/src/tokenizer.js
xtuc: I think we can add an error state, but the perf concern is valid. We can leave it as is for now :+1:

packages/wast-parser/src/fsm.js
xtuc: Fixed in https://github.com/xtuc/webassemblyjs/pull/215/commits/9b3f05b375dbeac2e80572525d467e604ee96cbb.

`n` and `allowedSeperator` are optional keys, it's `n?: number` in flow. `n: ?number` means that the key is needed but can be undefined/null.

xtuc:
> But parsing shouldnt be done entirely in the tokenize function imo.

Yes I agree.

_

It would be great to re-use the FSM class in the grammar and in the wasm parser in the future. Do you mind if I create a new package `@webassemblyjs/helper-fsm`?

maurobringolf:
Nice flow fix!

The fsm is not really webassembly specific, so Id rather publish it standalone instead. But I would like to solve the input copying problem first.

maurobringolf:
lgtm, do you need anything from me? We can merge I think

PR #220, opened 2018-03-29
Wasm parser testing

ColinEberhardt:
This PR cherry picks a few of the commits I removed from #212 (merged). The main focus here is testing the wasm decoder, this is done via wat files that are converted to wasm via wabt. I'm slowly replicating the wat-parser tests, although minor modifications are required as the wasm-parser tests cannot be fragments.

The commit https://github.com/xtuc/webassemblyjs/commit/68f74c924dd89e0683c5fd486484eb62aefa7d96 is a relatively big one as it involved regeneration of a number of expected files - as I've harmonised the way unique names are generated.

I expect to find a lot more issues as I go. I'll fix simple ones such as https://github.com/xtuc/webassemblyjs/commit/a618bdabe71c1c4889fce953dc1f9379d3ac809e (adding return types to blocks), and will raise issues for more complex ones.


packages/wasm-parser/src/decoder.js
```diff
@@ -62,16 +62,7 @@ function byteArrayEq(l: Array<Byte>, r: Array<Byte>): boolean {
...
-
-  function getUniqueName(prefix: string = "temp"): string {
-    if (!inc.hasOwnProperty(prefix)) {
-      inc[prefix] = 0;
-    } else {
-      inc[prefix] = inc[prefix] + 1;
-    }
-    return prefix + "_" + inc[prefix];
-  }
+  const getUniqueName = t.getUniqueNameGenerator();
```
xtuc: Looks cleaner :+1:

packages/wasm-parser/test/fixtures/block/with-instructions/actual.wat
```diff
@@ -0,0 +1,10 @@
+(module
+  (func
+    (block
+      (nop)
+      (nop)
+    )
+  )
+)
+
+
```
xtuc: Nit: could you please remove the extra nl here?

packages/wast-parser/src/grammar.js
```diff
@@ -448,7 +442,7 @@ export function parse(tokensList: Array<Object>, source: string): Program {
      */
     function parseBlock(): BlockInstruction {
       let label = t.identifier(getUniqueName("block"));
-      let blockResult;
+      let blockResult = null;
```
xtuc: The default shoud probably be `0x40` it's null in the blockTypes

packages/ast/src/index.js
```diff
@@ -251,13 +251,15 @@ export function blockInstruction(
...
-  return {
+  const node: BlockInstruction = {
     type: "BlockInstruction",
     id: "block",
     label,
     instr,
     result
   };
+
+  return node;
```
xtuc: The return type is already `BlockInstruction`, didn't flow catch that?

packages/ast/src/index.js
```diff
@@ -332,6 +334,18 @@ export function numberLiteral(
   return x;
 }
 
+export function getUniqueNameGenerator(): string => string {
+  const inc = {};
+  return function(prefix: string = "temp"): string {
+    if (!inc.hasOwnProperty(prefix)) {
```
xtuc: `if (prefix in inc)` looks better?

packages/wasm-parser/test/index.js
```diff
@@ -5,52 +5,77 @@ const diff = require("jest-diff");
...
+  if (actual.trim() !== expected.trim()) {
     throw new Error("Assertion error");
   }
 }
 
 describe("compiler", () => {
   describe("Binary format parsing", () => {
     const testSuites = glob.sync(
-      "packages/wasm-parser/test/fixtures/**/actual.wasm"
+      "packages/wasm-parser/test/fixtures/block/**/actual.wat"
```
xtuc: Just a reminder to remove this later and technically they are `wast` files.

packages/ast/src/index.js
ColinEberhardt: Good spot - that is a hangover from a refactor / revert

packages/wasm-parser/test/index.js
ColinEberhardt: good spot - thanks

packages/wast-parser/src/grammar.js
ColinEberhardt: @xtuc just realised I didn't address this comment? Would you prefer ...

```
import {blockTypes }  from '@webassembljs/helper-wasm-bytecode';
...
let blockResult = blockTypes.0x40;
```

Feels like an odd dependency to me ... wast-parser depending on helper-wasm-bytecode - unless I misunderstood.

xtuc: You are right, let's keep it null

PR #223, opened 2018-03-30
ESLint plugin

xtuc:
This is a first try, we can check a lot more.

I'm targeting usage in Webpack and Rollup.

See README https://github.com/xtuc/webassemblyjs/blob/feat-eslint-plugin/packages/eslint/README.md#supported-rules


packages/eslint/README.md
```diff
@@ -0,0 +1,68 @@
...
+}
+```
+
+
+Then configure the rules you want to use under the rules section.
+
+```json
+{
+    "rules": {
+        "webassembly/non-unexisting-export": 2
```
ColinEberhardt: Should that be `no-unexisting-export`

packages/eslint/README.md
```diff
@@ -0,0 +1,68 @@
...
+    }
+}
+```
+
+## Supported Rules
+
+### `no-unexisting-export`
```
ColinEberhardt: I must admit, I'm not keen on that naming. How about `no-missing-exports`? or `no-exports-missing`?

packages/eslint/src/rules/no-unexisting-export.js
```diff
@@ -0,0 +1,99 @@
...
+
+        const dirname = path.dirname(context.getFilename());
+
+        const source = object.arguments[0].value;
+        const file = path.join(dirname, source);
+
+        if (existsSync(file) === false) {
+          return context.report({
+            node: object,
+            message: "WASM file does not exists"
```
ColinEberhardt: `WASM file does not exist`

packages/eslint/README.md
xtuc: I agree, what about `no-unknown-export`?

ColinEberhardt:
nice idea!

packages/eslint/README.md
ColinEberhardt: yeah, `no-unknown-export` sounds good too

xtuc:
Thanks for your review @ColinEberhardt, I updated this.

PR #225, opened 2018-04-01
fix: module export references should be identifiers

### LINKED ISSUE #218: different AST produced for ModuleExport from wasm / wat parser
Given the following wat:

```
(export "test" (func 1))
```

The WAT parser creates this AST:

```
          {
            "type": "ModuleExport",
            "name": "test",
            "descr": {
              "type": "Func",
              "id": {
               "type": "NumberLiteral",
               "value": 1,
               "raw": "1"
              }
            }
          }
```

However, converting this wat to wasm, via wabt, then parsing with th ewasm-parser, the following AST is generated:

```
          {
            "type": "ModuleExport",
            "name": "test",
            "descr": {
              "type": "Func",
              "id": {
               "type": "Identifier",
               "value": "func_2"
              }
            }
          }
```

Which is correct?

The wasm parser output looks more logical to me!

ColinEberhardt:
fixes #218


packages/webassemblyjs/src/interpreter/runtime/values/module.js
```diff
@@ -292,39 +292,13 @@ function instantiateExports(
   traverse(n, {
     ModuleExport({ node }: NodePath<ModuleExport>) {
       if (node.descr.type === "Func") {
-        // Referenced by its index in the module.funcaddrs
-        if (node.descr.id.type === "NumberLiteral") {
```
xtuc: Could we maybe throw here? like "internal compiler failure". Because if we get into this it will be very hard to debug.

ColinEberhardt: Yeah, good idea - I've done a slightly more aggressive refactor in https://github.com/xtuc/webassemblyjs/pull/225/commits/942797a8e62e31d594b3dd85d1664d3d789fe9f9 so that all the temporary 'instantiated' items have the same structure. This means I was able add this exception logic in a single place.

packages/wast-parser/test/fixtures/export/global/expected.json
```diff
@@ -37,9 +37,8 @@
           "descr": {
             "type": "Global",
             "id": {
-              "type": "NumberLiteral",
-              "value": 0,
-              "raw": "0"
+              "type": "Identifier",
+              "value": "global_0"
```
xtuc: We should preserve the `raw` value because it will be printed as `global_0` otherwise.

ColinEberhardt: Good point, I presume the raw values are used for round tripping back to wasm / wat?

xtuc: I'm not sure what do you mean by "round tripping back to wasm / wat" but we already have deterministic transformation for Identifier ⟶ Indice and the inverse. We can transform for wasm as well as for wast.

`raw` is just for printing I think.

ColinEberhardt: I think this is a bigger change, so I've pulled it out into this issue for clarification https://github.com/xtuc/webassemblyjs/issues/228

PR #227, opened 2018-04-02
Move hex float parser

### LINKED ISSUE #226: Move the webassembly-floating-point-hex-parser package here
As discussed with @maurobringolf we can move the `webassembly-floating-point-hex-parser` package in the monorepo.

xtuc:
Closes #226


packages/floating-point-hex-parser/test/fuzzing/parse.c
```diff
@@ -0,0 +1,24 @@
...
+#include "time.h"
+#include "math.h"
+#include "float.h"
+
+int main(int argc, char *argv[]) {
+	// test id
+	int i = atoi(argv[1]);
+
+	/* Intializes random number generator */
+	srand(time(NULL) + i);
```
xtuc: @maurobringolf you were already passing an id from the test runner, I added it to the time to ensure no collisions (might not be the best solution tho). Do you know what time precision it uses? In my test i looked like seconds.

maurobringolf: Seems to be seconds, so this makes a lot of sense :+1: I can only guess from a quick google that it is not possible to get a more exact timestamp using C only: https://stackoverflow.com/a/8558644/6374865

xtuc:
Thanks @maurobringolf

PR #229, opened 2018-04-03
[WIP] better Stack validation and new package

maurobringolf:
I've had some time to look into validation of ast's, more specifically into type checking. This is a prototype that type checks a subset of all instructions and opts out of anything that is not yet supported, so it should never throw errors because something is not yet supported.

The idea is to run every function instruction per instruction and check if the types work out. In addition to that there is a `moduleContext` that maps ids to functions, globals, memories etc. It also contains information about the current stack frame like labels and locals. I couldn't find a reason to split the two, although they are conceptually very different (function scoped and module scoped).

There is a lot of work left to do, but the following testcases illustrate what type errors are already being caught correctly:

https://github.com/xtuc/webassemblyjs/tree/stack-validation-prototype/packages/webassemblyjs/test/fixtures/type-errors

@ColinEberhardt I already discussed a bit over Slack with Sven (should probably have used the public channel, sorry!) but if you have any inputs/suggestions/concerns I am happy to take them :wink: 

*(The code contains some weird checks and corner cases due to inconsistencies in the AST format, I'll make issues about them)*.

## TODO

- [x] `end` pseudo-instruction

- [x] Tables (nope, after merge --> Issue)

- [x] Decide what other validation passes to combine/integrate

- [x] Remove function parameters from initial stack

- [x] Assert that function results in *only* the expected result and not just the expected result on top of other things

- [x] Add test cases that run it over wasm

- [x] `unreachable`

- [x] Traps? Not sure how they relate to typing

- [x] Memory loads and stores plus its context

- [x] Complete imports ( currently just globals are implemented)

- [x] Find a better test format (that doesnt use 150 files)

- [x] `memory.grow` and `memory.size`

- [x] `trap`

- [x] Split into separate package

- [x] Add tests for invalid label values

- [x] Support  branches to all allowed constructs, currently only  block and loop are supported. Func, if etc also need to be.

- [x] Labels need to be removed from the module context when their blocks are left.


packages/webassemblyjs/test/fixtures/type-errors/add-mixed-operators/exec.tjs
```diff
@@ -0,0 +1,6 @@
+it("should throw an error during instantiation when type checking fails", () => {
+  assert.throws(
```
xtuc: Maybe we can take advange of the `throws.txt` we have for tokensizer. It has less noise that a entire test.

packages/webassemblyjs/test/index.js
```diff
@@ -46,10 +47,15 @@ describe("interpreter", () => {
...
   describe("wat", () => {
-    const testSuites = glob.sync(
+    let testSuites = glob.sync(
       "packages/webassemblyjs/test/fixtures/**/module.wast"
     );
 
+    // Filter based on last command line argument
+    if (argv.only) {
+      testSuites = testSuites.filter(f => f.includes(argv.only));
+    }
```
xtuc: I added a way to select tests, see [You can select test based on their titles](https://github.com/xtuc/webassemblyjs/blob/master/CONTRIBUTING.md#running-lintingtests). Would that fix this?

packages/webassemblyjs/test/fixtures/type-errors/add-mixed-operators/exec.tjs
maurobringolf: Ah yep I didnt consider this. Would have to modifiy the index.js accordingly

packages/webassemblyjs/test/index.js
maurobringolf: Ahh maybe I should have listened to GitHub when it told me "CONTRIBUTING.md has changed..." :grin: Thanks!

packages/webassemblyjs/src/compiler/validation/stack.js
```diff
@@ -0,0 +1,444 @@
...
+  return errors;
+}
+
+function applyInstruction(moduleContext, stack, instruction) {
+  // Return was called, skip everything
+  if (stack.return) {
+    return stack;
+  }
+
+  // Workaround for node.args which sometimes does not contain instructions (i32.const, call)
```
ColinEberhardt: I've spotted this too. The shape of the AST is something we can change - should these instructions be changed for consistency?

packages/webassemblyjs/src/compiler/validation/stack.js
```diff
@@ -0,0 +1,444 @@
...
+    case "trunc":
+    case "nearest": {
+      args = [instruction.object];
+      result = [instruction.object];
+      break;
+    }
+    /**
+     * binop
+     *
+     * @see https://webassembly.github.io/spec/core/valid/instructions.html#valid-binop
```
ColinEberhardt: It would be really cool if the validation rules could be generated from the specification - I do wonder whether that approach would be less effort in the long run?

ColinEberhardt:
> @ColinEberhardt I already discussed a bit over Slack with Sven (should probably have used the public channel, sorry!) but if you have any inputs/suggestions/concerns I am happy to take them 

No worries - looks good to me, just a couple of minor comments.

packages/webassemblyjs/test/fixtures/type-errors/add-mixed-operators/exec.tjs
xtuc: Well it's not easy actually, this packages uses the `exec.js` to instantiate the module. Let's leave it as is for now.

packages/webassemblyjs/src/compiler/validation/stack.js
maurobringolf: Interesting idea. Possibly for some that dont interact with the context (most them are already here though). The type rules that depend or modify the context (add label, change local etc) would be quite difficult I imagine

packages/webassemblyjs/src/compiler/validation/stack.js
maurobringolf: I made a list of possible ast modifications, Ill post them as separate issues today so we can decide individually

packages/webassemblyjs/src/compiler/validation/stack.js
maurobringolf: Some rules have type variables though, so a generator would have to understand the premises and generate code to load types from the context accordingly (load type of a local for example).

We need a translation from the type rules into the moduleContext API which seems difficult to automate for me.

Also note that all modifications of the context dont happen here but in "applyInstruction", so this list is not even the full type rules in those cases.

packages/webassemblyjs/src/compiler/validation/stack.js
```diff
@@ -523,6 +523,48 @@ function getType(moduleContext, stack, instruction) {
...
+    case "convert_u/f64":
+    case "reinterpret/f64": {
+      args = ["f64"];
+      result = [instruction.object];
+      break;
+    }
+    case "extend_u/i32":
+    case "extend_s/i32":
+    case "convert_s/i32":
+    case "convert_u/i32":
```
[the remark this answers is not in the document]
xtuc: Yeah, I think that we already doing it for the wast format anyway

xtuc:
I think we should figure out something for the tests, It seems that you added 150 files just for the tests. I won't probably have the time to fix, could you please take a look at it? Thanks!

maurobringolf:
I've added it to the todo list :+1: I'm pretty much done with the type rules themselves so I can focus on organizing the code and tests a bit better next.

ColinEberhardt:
Regarding tests, I do like the one where you can modify the test to regenerated the expected JSON. For others, I do feel like it would be worth merging.

packages/webassemblyjs/test/index.js
```diff
@@ -5,14 +5,43 @@ const chai = require("chai");
...
+        it("should give correct list of type errors", () => {
+          const ast = parse(module);
+          const errors = validate(ast);
+
+          assert.deepEqual(errors, expectedErrors);
+        });
+      });
+    });
+  });
+
```
maurobringolf: @xtuc What do you think of this way of running tests? See last two files in the diff for an exampl

ColinEberhardt: I'm not sure there is any value in having the test fixture, test data and expected output in three different places.

maurobringolf: Good suggestion, that makes sense for these small "explanatory" test cases :+1: I keep wondering if writing a parser for the spec tests `assert_invalid` construct is something we should do. I already started once but somehow regressed back to reformatting them by hand.

maurobringolf: We couldn't use the reformatter with this strategy though :thinking:

xtuc: That's a good point! The Try/catch + assert.equal logic already exists for this in the REPL.

Like:
```wat
(assert_invalid
  (module
    (func (result f32) (f32.convert_s/i32 (i64.const 0)))
  )
  "Expected type i32 but got i64"
)
```

Note that currently `make test` doesn't run the spec tests you will need to add this (maybe just your tests).

maurobringolf: I haven't looked into the repl in detail, but combined some more files into single test cases. The build still breaks due to some issue with the spec testsuite though. Locally my `testsuite` directory is empty is that correct? Even `master` doesnt pass `make-ci` locally for me, so I must be doing something wrong.

xtuc: You need to update your gitsubmodules

xtuc: I'll review it shortly, I think we have more than a MVP because it's already better than my first implementation.

xtuc:
Since https://github.com/xtuc/webassemblyjs/pull/250 is merge, we can add the code frame in the errors :+1:

packages/validation/test/fixtures/br/_module.wast
```diff
@@ -0,0 +1,25 @@
+(module
```
xtuc: this test is disabled right?

maurobringolf: Damn I forgot to delete this one, sorry! Good catch

xtuc:
We can remove this https://github.com/xtuc/webassemblyjs/blob/stack-validation-prototype/packages/validation/src/mut-global.js

packages/validation/src/type-checker/module-context.js
```diff
@@ -0,0 +1,138 @@
...
+    this.funcs.unshift({ args, result });
+  }
+
+  hasFunction(index) {
+    return typeof this.getFunction(index) !== "undefined";
+  }
+
+  getFunction(index) {
+    if (typeof index !== "number") {
+      throw new Error("getFunction only supported for number index");
```
xtuc: When name are restored from the binary, getFunction is called with an Identifer and the lookup doesn't work. Maybe we should disable name resolution or keep both.

xtuc:
@maurobringolf only errors left are unsupported `else` instruction, let's merge this after it

maurobringolf:
Awesome thank you for helping with the annoying fixes!!

Issue #184, opened 2018-02-26
Fix cycle in the dependency graph


xtuc:
It has been introduced by #180 

```
lerna WARN ECYCLE Encountered a cycle in the dependency graph.This may cause instability! Packages in cycle are: "@webassemblyjs/ast", "@webassemblyjs/cli", "@webassemblyjs/dce", "@webassemblyjs/wasm-gen", "@webassemblyjs/wasm-parser", "@webassemblyjs/wast-parser", "@webassemblyjs/wast-printer", "webassemblyjs"
lerna WARN ECYCLE Encountered a cycle in the dependency graph.This may cause instability! Packages in cycle are: "@webassemblyjs/ast", "@webassemblyjs/cli", "@webassemblyjs/dce", "@webassemblyjs/wasm-gen", "@webassemblyjs/wasm-parser", "webassemblyjs"
lerna WARN ECYCLE Encountered a cycle in the dependency graph.This may cause instability! Packages in cycle are: "@webassemblyjs/ast", "@webassemblyjs/cli", "@webassemblyjs/dce", "webassemblyjs"
```

PR #230, opened 2018-04-03
More wasm parser tests

ColinEberhardt:
A few small changes - removing a few `expected.json` files that shouldn't have been there (wasm-parser tests don't use them), and adding `br_if` / `br_table` tests. For `br_table` found a small bug in wasm-parser, fixed in this commit https://github.com/xtuc/webassemblyjs/commit/babaaea48cfa544e9f8955c6453d856f88c61424


packages/wasm-parser/src/index.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import { traverse } from "@webassemblyjs/ast";
+const t = require("@webassemblyjs/ast");
```
xtuc: I forgot to warn you, since we have a cyclic dependency with some modules, JavaScript Module sometime doesn't work, but I guess you already figure that out. There's an issue that I need to fix about it #184

xtuc:
Cool good catch! thanks

ColinEberhardt:
no problem - is this good to merge / approved?

PR #234, opened 2018-04-04
Fix anynomous identifiers

xtuc:
This adds `identifier.raw = ""` when it was anonymous and in the printing we ignore them.

Note that empty string is not a valid identifier.

Relevant test change https://github.com/xtuc/webassemblyjs/pull/234/files#diff-31198a6e4943bb5ba200294747816eb4


xtuc:
Yes this commit https://github.com/xtuc/webassemblyjs/pull/234/commits/ad58aeb327e3807b5fbb980e43eff38e65882790 is fixing the issue.

xtuc:
That's right, but the `raw` field is not mandatory (if value === raw we don't need it). I can be consistency with location handling: `withLoc`. What do you think about a `withRaw` function?

xtuc:
In my idea and how it's currently done. `raw` is an optional member of identifier; it's generic enough to support memory or table using the same paths.

I'll merge this and I hope that I addressed your concern, otherwise we can do a followup pr.

PR #241, opened 2018-04-06
doc: document AST design

ColinEberhardt:
I keep making reference to these unwritten rules - so thought it was worth writing them down!

### DIFF
--- a/docs/contrib-ast.md
+++ b/docs/contrib-ast.md
@@ -1,0 +1,26 @@
+
+---
+title: AST Design
+id: contrib-ast
+---
+
+The AST is used by a number of webassemblyjs packages as a way of representing wasm modules. It is produced / consumed by the parsers / printers that handle the binary WASM and textual WAST formats:
+
+```
+WASM  → wasm-parser ↘     ↗ wasm-gen     → WASM
+                      AST
+WAST  → wast-parser ↗     ↘ wast-printer → WAST
+
+```
+
+The structure of this AST is enforced via flow typing, and the `@webassemblyjs/ast` package that provides factory method for constructing AST nodes of the correct shape.
+
+There are some significant differences between the WASM and WAST representation of modules that have an impact on the AST:
+ - WASM doesn't have to include function and label names. This information may be supplied as a custom name section.
+ - WAST representation has some redundancy, function signatures can be specified 'inline', whereas with WASM function signatures are represented in a section and referenced by index.
+
+A design goal of our AST and the associated parsers / printers is that semnatically equivalent WASM and WAST files should result in an identical AST representation. This results in a simpler interpreter that does not need to handle various AST representations of the same 'application'.
+
+There are some exceptions to this rule ...
+
+With WAST it is possible to hve multiple representations of the same value, e.g. a constant value of`10000` is the same as `10_000`. In order to preserve this information, certain AST nodes have a `raw` property.
--- a/docs/contrib-ast.md
+++ b/docs/contrib-ast.md
@@ -4,7 +4,7 @@ title: AST Design
 id: contrib-ast
 ---
 
-The AST is used by a number of webassemblyjs packages as a way of representing wasm modules. It is produced / consumed by the parsers / printers that handle the binary WASM and textual WAST formats:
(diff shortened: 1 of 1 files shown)


docs/contrib-ast.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+
+The structure of this AST is enforced via flow typing, and the `@webassemblyjs/ast` package that provides factory method for constructing AST nodes of the correct shape.
+
+There are some significant differences between the WASM and WAST representation of modules that have an impact on the AST:
+ - WASM doesn't have to include function and label names. This information may be supplied as a custom name section.
+ - WAST representation has some redundancy, function signatures can be specified 'inline', whereas with WASM function signatures are represented in a section and referenced by index.
+
+A design goal of our AST and the associated parsers / printers is that semnatically equivalent WASM and WAST files should result in an identical AST representation. This results in a simpler interpreter that does not need to handle various AST representations of the same 'application'.
```
maurobringolf: Small typo here "semantically"

maurobringolf:
Very good idea, this will become more important when we do the flattening of nested instructions for example :+1:

docs/contrib-ast.md
```diff
@@ -0,0 +1,26 @@
+
+---
+title: AST Design
+id: contrib-ast
+---
+
+The AST is used by a number of webassemblyjs packages as a way of representing wasm modules. It is produced / consumed by the parsers / printers that handle the binary WASM and textual WAST formats:
```
xtuc: > the AST is our internal representation of any WebAssembly module in the project.

docs/contrib-ast.md
```diff
@@ -0,0 +1,26 @@
...
+WASM  → wasm-parser ↘     ↗ wasm-gen     → WASM
+                      AST
+WAST  → wast-parser ↗     ↘ wast-printer → WAST
+
+```
+
+The structure of this AST is enforced via flow typing, and the `@webassemblyjs/ast` package that provides factory method for constructing AST nodes of the correct shape.
+
+There are some significant differences between the WASM and WAST representation of modules that have an impact on the AST:
+ - WASM doesn't have to include function and label names. This information may be supplied as a custom name section.
```
xtuc: > This information may be supplied as a custom name section or randomly generated.

docs/contrib-ast.md
```diff
@@ -0,0 +1,26 @@
...
+ - WAST representation has some redundancy, function signatures can be specified 'inline', whereas with WASM function signatures are represented in a section and referenced by index.
```
xtuc: I guess it's just for documentation purposes, because we are planning to remove it via https://github.com/xtuc/webassemblyjs/issues/242 right?

docs/contrib-ast.md
xtuc: That's correct, the downside is that we are not able to re-print the same source. We can leave it for now, I'll think about a better solution.

docs/contrib-ast.md
```diff
@@ -0,0 +1,26 @@
...
+
+A design goal of our AST and the associated parsers / printers is that semnatically equivalent WASM and WAST files should result in an identical AST representation. This results in a simpler interpreter that does not need to handle various AST representations of the same 'application'.
+
+There are some exceptions to this rule ...
+
+With WAST it is possible to hve multiple representations of the same value, e.g. a constant value of`10000` is the same as `10_000`. In order to preserve this information, certain AST nodes have a `raw` property.
```
xtuc: > it possible to have multiple

xtuc: and a space after `of` and before `10000`. If we have the correct location information attached to that node (which I don't think we have) you can compare nodes.

docs/contrib-ast.md
ColinEberhardt: I really should get round to installing a VSCode spell check plugin!

ColinEberhardt: That's a very good point - when tackling #242 it might be worth separating out the normalisation from the parsing, so that re-printing becomes an option.

ColinEberhardt:
Yes, that would be good thanks

PR #246, opened 2018-04-11
Linear call stack

xtuc:
The `end` instruction is not widely used, it termantes the execution of the active frame.


xtuc:
I would like to know what do you think about the macros I introduced here? It's quite close to C/C++. Do you have feature requests? I'm missing optional params and macro-in-macro expansion.

Issue #242, opened 2018-04-06
The WAST parser should normalise function signatures


ColinEberhardt:
The current AST is de-normalized, in that the signature information is included in the function and call_indirect nodes.

With WASM the signatures / types are stored in a section and referenced by index. As a result, the wasm-parser currently de-normalises, by copying type information into the function nodes.

In order to ensure a standard and simple AST form (see #241), the WAST parser should normalise, storing the type information separately, and referencing from the func / call_indirect.

As a result, the AST generated from parsing WAST where the signature is inline:

```
(func (result i32)
  (i32.const 5)
)
```

Would be normalised to a form that is more closely aligned with the following:

```
(type (func (result i32)))
(func (type 0) 
  (i32.const 5)
)
```

Which is itself more closely aligned with the section-based WASM structure

ColinEberhardt:
With the changes for #257 function AST nodes can now either include the signature directly, or indirectly as a type references. Currently the interpreter will throw a runtime exception if it encounters a type references.

PR #256, opened 2018-04-13
wast parser supports type references for func and call_indirect

ColinEberhardt:
just putting in some of the groundwork for #242 ...


types/AST.js
```diff
@@ -260,8 +260,10 @@ type Func = {
   // Only in WAST
   name: ?Index,
 
-  params: Array<FuncParam>,
-  result: Array<Valtype>,
+  params?: Array<FuncParam>,
+  result?: Array<Valtype>,
+  typeRef?: Identifier | NumberLiteral,
```
ColinEberhardt: Is there a better type to use for this? If not, perhaps we should define a type for this ...

```
type TypeRef = Identifier | NumberLiteral
```

types/AST.js
```diff
@@ -346,6 +348,8 @@ type CallIndirectInstruction = {
   // WAST
   params?: Array<FuncParam>,
   results?: Array<Valtype>,
+  typeRef?: Identifier | NumberLiteral,
```
ColinEberhardt: I'd really like to enforce the constraint that a node has ether `params` & `result` properties, or `typeRef` - is this possible?

types/AST.js
xtuc: Yes, there's `Index` https://github.com/xtuc/webassemblyjs/blob/1e06bbd2b3a6117ceca12efc7ed99ef652e4b741/types/AST.js#L46-L54

types/AST.js
xtuc: Yes with a variant type. Maybe something like:

```js
type ParamsAndResults = {
   params?: Array<FuncParam>,
   results?: Array<Valtype>
};

type TypeRef = TypeInstruction | ParamsAndResults

type CallIndirectInstruction = {
  ref: TypeRef
};
```

ColinEberhardt: I've expanded on this and outlined in an issue: https://github.com/xtuc/webassemblyjs/issues/257

ColinEberhardt:
This PR has been updated to fix #257 by adding a new `Signature` type.

packages/webassemblyjs/src/interpreter/kernel/exec.js
```diff
@@ -234,9 +234,14 @@ export function executeStackFrame(
          */
         if (typeof func.name === "object") {
           if (func.name.type === "Identifier") {
+            if (func.signature.type !== "Signature") {
+              newRuntimeError(
```
xtuc: Just noticed, you're missing a `throw` before. It returns an error.

PR #264, opened 2018-04-17
Add validations package

xtuc:
Moved the validations into their own package.


ColinEberhardt:
Minor point - I'd call this package 'validation' rather than 'validations'. Otherwise, looks good to me!

xtuc:
Ok @ColinEberhardt I renamed it. I should have waited before publishing it (https://www.npmjs.com/package/@webassemblyjs/validations will remain for ever).

xtuc:
Waiting for your review @maurobringolf since it's against your branch.

maurobringolf:
Thank you! I also prefer the name validation, closer to the spec :+1:

PR #267, opened 2018-04-17
[wip] proper interpreter error

xtuc:
Relevant test in https://github.com/xtuc/webassemblyjs/tree/2f4a13fb43012d105ded4b23b28e09a0f66ba89a/packages/webassemblyjs/test/fixtures/error/add:

```wat
(module
  (func (result i32)
    (i32.const 1)
    (i32.add)
    ^^^^^^^^
    expected c1 value of type i32 on top of the stack, given type: label

    (end)
  )
  (export "add" (func $func_0))
)
```


packages/webassemblyjs/src/interpreter/kernel/exec.js
```diff
@@ -240,7 +246,10 @@ export function executeStackFrame(
     }
 
     function newRuntimeError(msg) {
-      return new RuntimeError(msg);
+      const e = new RuntimeError(msg);
+      e._appstack = stack;
```
xtuc: I'm not a super fan of that. The Error object is unspecified and I hope that this will always work.

xtuc:
Note to myself; store the source string instead of the AST? Would avoid the printing.

PR #269, opened 2018-04-18
AST transform for denormalising type references

ColinEberhardt:
When compiling a module, this AST transform denormalises type references so that func / call_indirect instructions have inline signatures.

As a result, the interpreter now supports the following:

```
(module
  (type $i32return (func (result i32)))
  (func (type $i32return) (i32.const 1))
)
```


maurobringolf:
Really cool!

Do I read this correctly that you dont remove the type instruction from the ast? Reading diffs on mobile is hard 😄

ColinEberhardt:
> Do I read this correctly that you dont remove the type instruction from the ast?
Correct -  leave them as-is.

packages/ast/test/transform/denormalize-type-references/index.js
```diff
@@ -0,0 +1,51 @@
...
+  }
+  if (a.trim() !== b.trim()) {
+    throw new Error("Assertion error");
+  }
+};
+
+const loadJSON = filename =>
+  JSON.parse(readFileSync(path.join(__dirname, filename), "utf8"));
+
+const copy = json => JSON.parse(JSON.stringify(json));
```
xtuc: I think that we can use `const copy = o => Object.assign({}, o);` here.

ColinEberhardt: I think that will just perform a shallow copy?

xtuc: That's right. I was a bit worried about perf but it's in the test so doesn't matter so much :+1:

PR #270, opened 2018-04-20
validation package

xtuc:
Removing some noise from #229


maurobringolf:
This is just to include the package first, before adding the type checker right? I continue in the other branch for now

xtuc:
yes, and then I will update your branch/PR so we can get rid of the some useless changes.

PR #271, opened 2018-04-21
refactor: use test helper part1

xtuc:

PR #272, opened 2018-04-28
refactor(webassemblyjs): use test framework

xtuc:
Following #271


xtuc:
Nice to have but not priority now

PR #273, opened 2018-04-29
Fix webpack wasm edit bug

xtuc:
When using wasm-edit we assume that the LEB128 encoding (section size or section vector size) has the same number of bytes than our implementation. We found a case where they always padded it to 5 bytes and we assumed it was 1 byte.

This PR aims to preserve the padding. 

for ref https://github.com/webpack/webpack/pull/7144

### DIFF
--- a/packages/wasm-parser/src/decoder.js
+++ b/packages/wasm-parser/src/decoder.js
@@ -318,26 +318,10 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
       if (type == types.func) {
         dump([type], "func");
 
-        const paramValtypes: Array<Valtype> = parseVec(b => {
-          const t = valtypes[b];
-
-          if (typeof t === "undefined") {
-            throw new Error("Unknown func parameter valtype: " + toHex(b));
-          }
-
-          return t;
-        });
+        const paramValtypes: Array<Valtype> = parseVec(b => valtypes[b]);
         const params = paramValtypes.map(v => t.funcParam(v));
 
-        const result: Array<Valtype> = parseVec(b => {
-          const t = valtypes[b];
-
-          if (typeof t === "undefined") {
-            throw new Error("Unknown func result valtype: " + toHex(b));
-          }
-
-          return t;
-        });
+        const result: Array<Valtype> = parseVec(b => valtypes[b]);
 
         const endLoc = getPosition();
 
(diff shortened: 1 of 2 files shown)


packages/helper-wasm-section/src/resize.js
```diff
@@ -33,7 +33,29 @@ export function resizeSectionByteSize(
...
+
+  const oldu32EncodedLen = end - start;
+  const newu32EncodedLen = newBytes.length;
+
+  // For padded LEB128 we can end up resizing the section again because we don't
+  // use the same encoding (thus number of bytes)
+  if (newu32EncodedLen !== oldu32EncodedLen) {
+    const deltaInSizeEncoding = newu32EncodedLen - oldu32EncodedLen;
+
+    sectionMetadata.size.loc.end.column = start + newu32EncodedLen;
```
sendilkumarn: wont the end be, `end + newu32EncodedLen`

packages/wasm-edit/src/apply.js
```diff
@@ -192,19 +194,37 @@ function applyAdd(ast: Program, uint8Buffer: Uint8Array, node: Node): State {
   const newByteArray = encodeNode(node);
 
   // start at the end of the section
-  const start = sectionMetadata.startOffset + sectionMetadata.size.value + 1;
+  // the section size doesn't include the encoded u32 length (of the actual size)
+  // and of the vector size, we need to add them to the section size.
+  const u32EncodedLengths =
+    sectionMetadata.size.loc.end.column -
```
sendilkumarn: braces makes sense here? adding the difference right?

packages/wasm-parser/src/decoder.js
```diff
@@ -319,10 +318,26 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
       if (type == types.func) {
         dump([type], "func");
 
-        const paramValtypes: Array<Valtype> = parseVec(b => valtypes[b]);
+        const paramValtypes: Array<Valtype> = parseVec(b => {
+          const t = valtypes[b];
```
sendilkumarn: well this part can be a separate method I guess.

packages/helper-wasm-section/src/resize.js
xtuc: This is actually the end of the `size` of the section number.

packages/wasm-parser/src/decoder.js
xtuc: Yeah I agree, it wrote this to have a better error message when debuging

packages/wasm-edit/src/apply.js
xtuc: Yeah but Prettier removed them but math is still fine

PR #276, opened 2018-05-03
Fix tests in wasm-parser

xtuc:
I noticed that our `wasm-parser` tests were inactive.

The actual and expected JSON was:

```json
 {
  "body": [
    {
      "type": "Module"
    }
  ],
  "type": "Program"
}
```

It seems to be caused by the replacer.

cc @ColinEberhardt


xtuc:
@ColinEberhardt do you know where are block labels stored? Also in the custom name section?
  https://github.com/xtuc/webassemblyjs/blob/fed152017fd5d19c09c8c53e3369d849e7c538f2/packages/wasm-parser/test/fixtures/block/with-label/actual.wat-failure#L3

Issue #282, opened 2018-05-06
data text not shown in wast-printer


xtuc:
prints:
```wat
  (data 0 (i32.const 16) "")
```

where is should be:
```wat
  (data (i32.const 16) "test\00"))
```

PR #283, opened 2018-05-07
decode data in wasm

xtuc:
As said in the comment some binaries uses big chunks of data. Since we know the length in advance we could throw, also depends on the `ignoreDataSection` option. 

Closes https://github.com/xtuc/webassemblyjs/issues/282


xtuc:
I'm not able to reproduce the hangs when parsing big chunks of data. We can add a limit or disable printing later.

PR #286, opened 2018-05-08
Traverse refactor

### LINKED ISSUE #263: Simplify the AST traversal
I'm in the process of adding a new transform that denormalizes type references as per #242. It feels like more work than it should be, when you add a new node type, or node property you have to update the AST traversal to accommodate this structural change (adding extra traversal logic and node type information).

I'd like to refactor so that it traverses all object properties automatically, invoking the callback for all objects that have a `type` property. Can anyone see a problem with this?

Also, should the AST transforms be moved into the `@webassemblyjs/ast` package? I've added a few in wasm-parser (e.g. #262), I'd be happy to move them into a common package.

ColinEberhardt:
fixes #263 

This was surprisingly straightforward, the traverse function is now only around 10 lines of code. Basically it recurses the AST invoking the callback on any object that has a `type` property.

The only issue was the module exports, currently they have an AST representation that is a little misleading, e.g. a function export is as follows:

```
{
  "type": "ModuleExport",
  "name": "add",
  "descr": {
    "type": "Func",
    "id": {
      "type": "Identifier",
      "value": "func_0"
    }
  }
}
```

The above looks like a `Func` node, when it isn't.

I've modified the AST to ass a `ModuleExportDesc` node type:

```
{
  "type": "ModuleExport",
  "name": "add",
  "descr": {
    "type": "ModuleExportDescr",  <---- new type
    "exportType": "Func",                <---- describes the type being referenced
    "id": {
      "type": "Identifier",
      "value": "func_0"
    }
  }
}
```

TODO:
 - [x] fix the one failing unit test
 - [x] add a bit of documentation about how traverse works

### DIFF
--- a/packages/ast/src/traverse.js
+++ b/packages/ast/src/traverse.js
@@ -93,212 +93,22 @@ function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
     return;
   }
 
-  switch (n.type) {
-    case "Program": {
-      const path = createPath(n, parentPath);
-      cb(n.type, path);
+  const path = createPath(n, parentPath);
+  // $FlowIgnore
+  cb(n.type, path);
 
-      n.body.forEach(x => walk(x, cb, path));
-
-      break;
+  Object.keys(n).forEach((prop: string) => {
+    const value = n[prop];
+    if (!value) {
+      return;
     }
-
-    case "SectionMetadata":
-    case "FunctionNameMetadata":
-    case "ModuleNameMetadata":
-    case "LocalNameMetadata":
-    case "Data":
-    case "Memory":
-    case "Elem":
-    case "FuncImportDescr":
-    case "GlobalType":
-    case "NumberLiteral":
-    case "ValtypeLiteral":
-    case "FloatLiteral":
-    case "StringLiteral":
-    case "QuoteModule":
-    case "LongNumberLiteral":
-    case "BinaryModule":
-    case "LeadingComment":
-    case "BlockComment":
-    case "Identifier": {
-      cb(n.type, createPath(n, parentPath));
-      break;
-    }
-
-    case "ModuleExport": {
-      const path = createPath(n, parentPath);
-      cb(n.type, path);
-
-      walk(n.descr.id, cb, path);
-
-      break;
-    }
-
-    case "ModuleMetadata": {
-      const path = createPath(n, parentPath);
-      cb(n.type, path);
-      n.sections.forEach(x => walk(x, cb, path));
-
-      if (typeof n.functionNames !== "undefined") {
-        // $FlowIgnore
-        n.functionNames.forEach(x => walk(x, cb, path));
+    const valueAsArray = Array.isArray(value) ? value : [value];
+    valueAsArray.forEach(v => {
+      if (v.hasOwnProperty("type")) {
+        walk(v, cb, path);
       }
-
-      if (typeof n.localNames !== "undefined") {
-        // $FlowIgnore
-        n.localNames.forEach(x => walk(x, cb, path));
-      }
-      break;
-    }
-
-    case "Module": {
-      const path = createPath(n, parentPath);
-      cb(n.type, path);
-
-      if (typeof n.fields !== "undefined") {
-        n.fields.forEach(x => walk(x, cb, path));
-      }
-
-      if (typeof n.metadata !== "undefined") {
-        // $FlowIgnore
-        walk(n.metadata, cb, path);
-      }
-
-      break;
-    }
-
-    case "Start":
-    case "CallInstruction": {
-      const path = createPath(n, parentPath);
-      // $FlowIgnore
-      cb(n.type, path);
-
-      // $FlowIgnore
-      walk(n.index, cb, path);
-
-      break;
-    }
-
-    case "CallIndirectInstruction": {
-      const path = createPath(n, parentPath);
-      // $FlowIgnore
-      cb(n.type, path);
-
-      if (n.index != null) {
-        // $FlowIgnore
-        walk(n.index, cb, path);
-      }
-
-      break;
-    }
-
-    case "ModuleImport": {
-      cb(n.type, createPath(n, parentPath));
-
-      if (n.descr != null) {
-        // $FlowIgnore
-        walk(n.descr, cb, createPath(n, parentPath));
-      }
-
-      break;
-    }
-
-    case "Table":
-    case "Global": {
-      const path = createPath(n, parentPath);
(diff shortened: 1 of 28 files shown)


packages/webassemblyjs/test/fixtures/call-func-with-i64-arg/exec.tjs
```diff
@@ -1,6 +1,7 @@
 it("should throw when calling the exported function", () => {
   const m = WebAssembly.instantiateFromSource(watmodule);
 
+  debugger;
```
ColinEberhardt: ooops - I'll remove that!

packages/ast/src/traverse.js
```diff
@@ -93,212 +93,22 @@ function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
...
-        n.init.forEach(x => walk(x, cb, path));
+  const path = createPath(n, parentPath);
+  // $FlowIgnore
+  cb(n.type, path);
+
+  Object.keys(n).forEach((prop: string) => {
+    const value = n[prop];
+    if (!value) {
+      return;
+    }
```
xtuc: I think this can could just be 
```js
if (prop === "type") {
   walk(n, cb, path);
} else if (Array.isArray(prop)) {
  n[prop].forEach(x => walk(x, cb, path))
}
```

xtuc:
I came up with 
```js
function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
  if (n._deleted === true) {
    return;
  }

  const path = createPath(n, parentPath);

  Object.keys(n).forEach((prop: string) => {
    const v = n[prop];

    if (prop === "type") {
      cb(n.type, path);
    } else if (Array.isArray(v)) {
      v.forEach(f => walk(f, cb, path));
    }
  });
}
```

but it's failing 3 tests.

ColinEberhardt:
> but it's failing 3 tests.

ColinEberhardt:
> but it's failing 3 tests.

Your logic doesn't quite look right to me - in my implementation `walk` invokes the callback for the current node, than iterates over all of its properties. For any property value (or child of an array property) that has a `type` property, it recurses.

Yours is doing something a bit different (that is hard to explain!)

xtuc:
Yes, at first I didn't quite understand your implementation, that's why I tried my own.

I think that in your implementation we can remove the first check (`!value`).

packages/ast/src/traverse.js
```diff
@@ -93,212 +93,22 @@ function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
...
+    const valueAsArray = Array.isArray(value) ? value : [value];
+    valueAsArray.forEach(v => {
+      if (v.hasOwnProperty("type")) {
```
xtuc: do you mind using `typeof v.type === "string"` instead here? Or do we need to check for own property explicitly?

packages/ast/src/traverse.js
```diff
@@ -93,212 +93,22 @@ function walk(n: Node, cb: Cb, parentPath: ?NodePath<Node>) {
...
-      if (n.init != null) {
-        // $FlowIgnore
-        n.init.forEach(x => walk(x, cb, path));
+  const path = createPath(n, parentPath);
+  // $FlowIgnore
+  cb(n.type, path);
+
+  Object.keys(n).forEach((prop: string) => {
+    const value = n[prop];
+    if (!value) {
```
xtuc: Following my comment, I think that this check is not needed. Also it could lead to confusing like with node like `{hasFoo: false}`.

ColinEberhardt:
Some good points, thanks for the feedback - I'll make some changes tomorrow.

packages/ast/src/traverse.js
ColinEberhardt: @xtuc this check is needed (although to your point on `false` is not implemented correctly), because currently the AST has a number of [maybe types](https://flow.org/en/docs/types/primitives/#toc-maybe-types). I do wonder whether these should be changed to [optional types](https://flow.org/en/docs/types/primitives/#toc-optional-object-properties) instead for consistency?

xtuc: Yes it's inconsistent because sometimes I wanted to always have a key even null. I don't mind if we switch to optional types instead.

But I think both can be covered with `value != null`.

ColinEberhardt:
@xtuc currently there are quite a few properties that are `undefined` at runtime:

TypeInstruction.id
Global.name
Module.id
Limit.max
LoopInstruction.result

I'd prefer to look at these as a separate PR, perhaps addressing the maybe / optional types. I also think the AST node constructor functions need a bit more validation, to enforce more consistency.

xtuc:
Yes, I agree.

We could maybe use some Babel work https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/jsx.js#L9-L26

The definitions are then used generates asserts/validations/builders: https://github.com/babel/babel/blob/master/packages/babel-types/scripts/generateTypeHelpers.js

ColinEberhardt:
> We could maybe use some Babel work https://github.com/babel/babel/blob/master/packages/babel-types/src/definitions/jsx.js#L9-L26

Nice idea - or maybe even generate the whole thing from the spec!

ColinEberhardt:
I've made the small changes request, hopefully good to merge?

packages/ast/src/traverse.js
```diff
@@ -2,7 +2,7 @@
 
 type Cb = (type: string, path: NodePath<Node>) => void;
 
-const debug = require("debug")("wasm:traverse");
+const debug = require("debug")("ast:traverse");
```
xtuc: I needed a way to grep for all our debug messages (excluding mocha ones for example), and I decided to pick the wrong `wasm` naming.

Maybe we could add `webassemblyjs:...` as prefix. I'll open an issue and send a PR.

scripts/test.sh
```diff
@@ -8,10 +8,10 @@ PACKAGES="./packages/*"
...
 
 ## Launch pretest scripts if provided
-for D in ./packages/*; do
-    (npm run pretest --silent --prefix $D || true) &
-    echo "launch pretest target for $D ..."
-done
+# for D in ./packages/*; do
+#     (npm run pretest --silent --prefix $D || true) &
+#     echo "launch pretest target for $D ..."
+# done
```
xtuc: This is annoying, i'll simplify that

ColinEberhardt:
Cool, that's merged. I'll start exploring a more metadata-driven approach to creating the AST type definitions and constructor functions.

xtuc:
Cool, thanks!

Issue #287, opened 2018-05-08
Change license to MIT


xtuc:
ref https://github.com/webpack/webpack/issues/7229

freund17:
I'm not a lawyer, but...
I'm pretty sure you need consent from every contributor in order to change the license.
Did you get that?

PR #288, opened 2018-05-08
re-license to MIT

xtuc:
Closes https://github.com/xtuc/webassemblyjs/issues/287


mohsen1:
Thanks! 

Please release those packages with a patch version so those license checkers become happy!

PR #300, opened 2018-05-10
Feat validation wasm

xtuc:


maurobringolf:
Some of the errors are coming from this https://github.com/xtuc/webassemblyjs/issues/302

packages/validation/test/index.js
```diff
@@ -40,6 +40,6 @@ describe("validation", () => {
       return errorsToString(errors);
     };
 
-    compareWithExpected(testSuites, pre, "throws.txt");
+    compareWithExpected(testSuites, pre, "output.txt");
```
xtuc: I would suggest keeping throws for parser/early errors since it's builin in the "test framework"

PR #301, opened 2018-05-10
Swap actual and expected in diffs

maurobringolf:


maurobringolf:
Yeah I looked here https://github.com/facebook/jest/blob/master/packages/jest-diff/src/index.js#L71 and from what I see it is really this way around

Issue #303, opened 2018-05-10
switch to yarn


xtuc:
Switch to yarn:
- [x] update CI
- [x] update CONTRIBUTING.md
- [x] remove and gitignore package-lock.json

PR #304, opened 2018-05-10
Fix remove lock files

xtuc:
Lockfiles are completely out-of-sync. Ref #303


xtuc:
Ok that doesn't work as expected

PR #305, opened 2018-05-10
AST node metadata generation

ColinEberhardt:
Following on from ideas discussed here: https://github.com/xtuc/webassemblyjs/pull/286#issuecomment-387626916

This is an early work-in-progress for generating the AST types and constructors functions from metadata.

Metadata is defined in `definitions.js`, which is then used to generate `types.js` and `constructorFunctions.js`

This gives a number of benefits:

- Consistency in naming, assertions etc ...
- Can be used to generate documentation
- Could potentially be used for runtime checking of AST
- It allows us to more easily change the patterns used, e.g. maybe vs optional types

Any thoughts / feedback?


xtuc:
Could you please add some indications for the generation?

types/AST.js
```diff
@@ -390,149 +224,11 @@ type CallIndirectInstruction = {
...
-
-type Memory = {
-  ...BaseNode,
-
-  type: "Memory",
-  limits: Limit,
-  id: ?Index
-};
-
+// TODO: any reason this is not a node?
```
xtuc: Not really

packages/ast/scripts/generateConstructors.js
```diff
@@ -0,0 +1,140 @@
...
+
+      ${assertParams(filterProps(typeDefinition.fields, f => !f.constant))}
+      ${buildObject(typeDefinition)} 
+
+      return node;
+    }
+    `;
+  });
+
+  fs.writeFileSync(filename, prettier.format(code));
```
xtuc: Instead of writing a file we could emit on the stdout, what do you think?

```
node scripts/generateConstructors.js | prettier > dist/something
```

ColinEberhardt:
@xtuc this feels like a good point to merge the first iteration. I've rebased to just a few commits.

The first introduces the metadata approach for generating node constructor functions and flow typing. The second commit is a small amount of refactoring to ensure that the rest of webassemblyjs uses these new constructor functions, removing the various inconsistencies and special cases.

I'm not sure of the best place to add the code gen to the build - any ideas?

After this, there is a bit more refactoring I'd like to do around the various `ObjectInstruction` / `Instruction` special cases. I'd also like to do some more tidyup on `index.js`, so that it simply becomes a file for exporting functionality from other modules.

packages/ast/README.md
```diff
@@ -51,9 +51,8 @@ console.log(signatures);
 - function `ifInstruction(testLabel, result, test, consequent, alternate)`
 - function `withLoc(n, end, start)`
 - function `moduleImport(module, name, descr)`
-- function `globalImportDescr(valtype, mutability)`
+- function `globalType(valtype, mutability)`
```
xtuc: This is auto generated by the `dump-export` package. Basically it lists every export.

xtuc:
and once again windows is failing for obscure reasons :(

packages/ast/README.md
ColinEberhardt: Ah - thanks for the heads-up. Eventually I'd like to generate this from flow documentation, giving a more rich API doc

ColinEberhardt:
> please let me know what do you think about my last changes. LGTM

Looks good to me - already making good use of the metadata!

I'm happy for you to merge whenever you're happy

PR #306, opened 2018-05-11
Push locals into function body

### LINKED ISSUE #302: Locals in wasm-parser
It seems like locals are not handled correctly in wasm-parser currently. Here is what I tried as a new testcase:

```
(module
  (func (result i64)
  	(local i32)
    (get_local 0)
  )
)
```

And here is what I get:

image

maurobringolf:
Fixes #302

### DIFF
--- a/packages/wasm-parser/test/index.js
+++ b/packages/wasm-parser/test/index.js
@@ -28,8 +28,8 @@ function stripMetadata(ast) {
   return ast;
 }
 
-// - Expected is wasm-parser
-// + Received is wast-parser
+// - Expected is wast-parser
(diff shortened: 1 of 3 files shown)


packages/wasm-parser/test/index.js
```diff
@@ -28,8 +28,8 @@ function stripMetadata(ast) {
   return ast;
 }
 
-// - Expected is wasm-parser
-// + Received is wast-parser
+// - Expected is wast-parser
+// + Received is wasm-parser
```
maurobringolf: This is because of #301

PR #308, opened 2018-05-11
Add more missing debug deps

christiango:
Some of the packages have a dependency on debug, but it isn't listed as a dependency in their package.json. This is causing issues for webpack users that use pnpm for package management since it results in the debug module not being able to be resolved.


xtuc:
Thanks and sorry!

PR #312, opened 2018-05-14
Else parsing for wasm

maurobringolf:
Currently else is always an empty sequence. This is a simplistic way of splitting the if body by an `else` instruction in it. If I understand the parsing procedure correctly, this should give the correct results for nested ifs as well (since inner else's are stripped of first).

This does fix all remaining errors in #229, I tried merging them locally.


packages/wasm-parser/src/decoder.js
```diff
@@ -709,15 +709,24 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
...
 
-        // FIXME(sven): handle the second block via the byte in between
-        const alternate = [];
+        const consequentInstr = ifBody.slice(0, elseIndex);
+        const alternate = ifBody.slice(elseIndex + 1);
 
         // FIXME(sven): where is that stored?
         // preserve anonymous
-        const testIndex = t.withRaw(t.identifier(getUniqueName("ifindex")), "");
+        const testIndex = t.withRaw(t.identifier(getUniqueName("if")), "");
```
maurobringolf: This did not break anything but fix a compatibility issue between wast-parser and wasm-parser, I havent looked into it in detail though.

maurobringolf:
There is some AST compatibility issue left with the `result` attribute on the `IfInstruction` node. I am not sure what is desired? @xtuc @ColinEberhardt 

image

xtuc:
Could we consistently have the `result`  prop?

PR #318, opened 2018-05-15
Add typechecker testsuite

xtuc:


Makefile
```diff
@@ -43,6 +43,7 @@ test-whitelisted-spec:
 	$(REPL) $(SPEC_TEST_DIR)/globals.wast
 	$(REPL) $(SPEC_TEST_DIR)/i32.wast
 	$(REPL) $(SPEC_TEST_DIR)/binary.wast
+	$(REPL) $(SPEC_TEST_DIR)/typecheck.wast --debug
```
xtuc: Remove the `--debug` when it's ready to be merge.

packages/repl/src/index.js
```diff
@@ -12,6 +12,7 @@ const {
 } = require("webassemblyjs/lib/interpreter/kernel/memory");
 const { decode } = require("@webassemblyjs/wasm-parser");
 const t = require("@webassemblyjs/ast");
+const typeCheck = require("@webassemblyjs/validation").stack;
```
maurobringolf: @xtuc I hope I didnt misunderstand things but to me it looked like no validation was running in the repl at all? I added it in a way that it is currently only run when using `assert_invalid` with `type mismatch` as expected error. As expected some case with `br` and `return` fails :grin:

packages/validation/src/type-checker.js
```diff
@@ -132,7 +133,7 @@ function checkStacks(expectedStack, actualStack) {
     }
 
     // There are still types left on the resulting stack
-    if (j >= 0) {
+    if (actualStack.slice(0, j + 1).filter(t => t !== POLYMORPHIC).length > 0) {
```
xtuc: Could we extract this into a new function? It's difficult to understand to me

maurobringolf: Sure, I'll make it more explicit. This is saying "If the stack contains anything else than polymorphic types". The reason is that a polymorphic type can also be `[]` so its valid in that case even though there are still types left on the stack after the function.

packages/repl/src/index.js
```diff
@@ -293,9 +282,19 @@ export function createRepl({ isVerbose, onAssert, onLog, onOk }) {
       _internalInstanceOptions: internalInstanceOptions
     };
 
-    const module = createCompiledModule(moduleNode);
+    if (enableTypeChecking === true) {
+      denormalizeTypeReferences(moduleNode);
```
maurobringolf: Ah this is much better, thanks :+1:

packages/repl/src/index.js
```diff
@@ -278,9 +281,20 @@ export function createRepl({ isVerbose, onAssert, onLog, onOk }) {
     const importObject = {
       _internalInstanceOptions: internalInstanceOptions
     };
-    const module = createCompiledModule(moduleNode);
 
-    return new Instance(module, importObject);
+    if (enableTypeChecking === true) {
+      denormalizeTypeReferences(moduleNode);
+
+      const typeErrors = typeCheck(t.program([moduleNode]));
```
xtuc: I just noticed that but type checking should be done an a module directly.

- Program can have multiple modules
- Program have no typing rules
- in fact I created the notion of Program

packages/repl/src/index.js
```diff
@@ -278,9 +281,20 @@ export function createRepl({ isVerbose, onAssert, onLog, onOk }) {
...
+
+      if (typeErrors.length > 0) {
+        throw new Error("type mismatch");
+      }
+    }
+
+    const compiledModule = createCompiledModule(moduleNode);
```
xtuc: See the implementation of `createCompiledModule`: https://github.com/xtuc/webassemblyjs/blob/master/packages/webassemblyjs/src/compiler/compile/module.js#L40-L43. 

It calls the root validation function which includes the new typechecker.

Why do we need to run the typechecker before again? It seems that there's a bug.

xtuc:
I just noticed that createCompiledModule takes a Program, I made to track/change it https://github.com/xtuc/webassemblyjs/issues/322.

PR #326, opened 2018-05-18
AST tidyup

ColinEberhardt:
This resolves all the TODOs added when generating the AST-related code


ColinEberhardt:
> But that's actually a breaking change right?

I guess it depends what we consider to be the public API? The AST is unchanged, however, the exported `numberLiteral` function has been renamed `numberLiteralFromRaw`, so that is breaking.

Issue #324, opened 2018-05-17
AST nodes lack line-of-code information


ColinEberhardt:
When adding WAT support to ast explorer https://github.com/fkling/astexplorer/pull/316 (merged and now live 🎉), I spotted that very few AST nodes actually contain line-of-code metadata.

PR #327, opened 2018-05-19
add locs

xtuc:
Fixes https://github.com/xtuc/webassemblyjs/issues/324


xtuc:
Ok this is far from being perfect but in a mergeable state now.

PR #333, opened 2018-05-21
Ensure section ordering in wasm

### LINKED ISSUE #280: Ensure section ordering in wasm
The order is defined here https://github.com/xtuc/webassemblyjs/blob/master/packages/helper-wasm-bytecode/src/index.js#L93-L106

We should throw an error (unexpected section x) when it's not meant to be the next sections. Custom section needs to be ignored.

sendilkumarn:
fixes #280


xtuc:
Looks cool!

Could you please add a test case? like:

```js
const b = makeBuffer(
  encodeHeader(),
  encodeVersion(1),
  [constants.sections.code, 0x01, 0x00],
  [constants.sections.func, 0x01, 0x00],
  [constants.sections.type, 0x01, 0x00]
);
```

This shouldn't decode since the sections are not in the correct order.

xtuc:
@sendilkumarn do you need help for the tests? I can do it if you want

PR #335, opened 2018-05-22
Feat identifier to index br

### LINKED ISSUE #332: The identifier-to-index transform doesn't convert br
Replace the label in `br` with the corresponding index.

Small example:

in

```wast
(module
  (func $label
    (br $label)
  )
)
```

out

```wast
(module
  (func $label
    (br 0)
  )
)
```

xtuc:
Closes #332

### DIFF
--- a/packages/ast/src/transform/wast-identifier-to-index/index.js
+++ b/packages/ast/src/transform/wast-identifier-to-index/index.js
@@ -1,6 +1,12 @@
 // @flow
 
-import { numberLiteralFromRaw, traverse, isIdentifier } from "../../index";
+import {
+  isBlock,
+  isFunc,
+  isIdentifier,
+  numberLiteralFromRaw,
+  traverse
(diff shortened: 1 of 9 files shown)


packages/ast/src/transform/wast-identifier-to-index/index.js
```diff
@@ -110,8 +126,48 @@ function transformFuncPath(
...
+
+        if (isIdentifier(firstArg) === true) {
+          // if the labels is not found it is going to be replaced with -1
+          // which is invalid.
+          let relativeBlockCount = -1;
+
+          // $FlowIgnore: reference?
+          instrPath.findParent(({ node }) => {
+            if (isBlock(node)) {
+              relativeBlockCount++;
```
xtuc: While this is not the best solution it's the safest. The alternative would be to add `enter` and `exit` hooks and simulate the label stack.

Issue #331, opened 2018-05-21
Define the `assert` macro project-wide


xtuc:
Currently macros are file specific but we commonly use the assert, we could declare it in the config.

PR #342, opened 2018-05-22
refactor: macro

xtuc:
Closes https://github.com/xtuc/webassemblyjs/issues/331


packages/ast/src/nodes.js
```diff
@@ -3,11 +3,11 @@
...
-function assert(cond: boolean) {
-  if (!cond) {
-    throw new Error("assertion error");
-  }
-}
+import { assert } from "mamacro";
+
+// FIXME(sven): compat with assert
+// eslint-disable-next-line no-unused-vars
+class RuntimeError extends Error {}
```
xtuc: The `assert` macro uses RuntimeError, which forces me to define it here. If i switch to `Error` into the macro many tests from the kernel/exec will fail, so I don't know what to do yet.

Maybe having two macros.

Issue #339, opened 2018-05-22
Allow instruction without parentheses


xtuc:
The following example should be parsed:

```wast
(func (i32.const 1) drop)
```

But we currently disallow the an instruction (`drop` here) without parentheses. 

This is fine:
```wast
(func (i32.const 1) (drop))
```

---

We need to use the instruction signatures to know when to stop to parse an instruction, for example:

```wast
(func i32.const 1 drop)
```

is valid too.

maurobringolf:
Ill have a go at this

xtuc:
We have a first support in https://github.com/xtuc/webassemblyjs/pull/343.

PR #343, opened 2018-05-22
First draft of parens-less instructions

maurobringolf:
I literally just tried to pass a few examples without breaking anything else. Im pretty sure this will need further tweaking.


xtuc:
Closes #339

xtuc:
I've fixed the flow errors on master, have you looked at them or ignored because of "references"?

Otherwise LGTM, we can merge an early version if you want.

maurobringolf:
I saw your fixes and rebased ontop of them, thanks for that. There are other errors now due to the additional vector property on `Array<Valtype>`. I am trying to come up with a nicer solution that flow agrees with.

xtuc:
Can we merge this? We can iterate on that later

maurobringolf:
Fine with me, I've had a go at the vector but couldnt get flow to agree.

xtuc:
ah no the CI :sob:

xtuc:
Thanks @maurobringolf

PR #344, opened 2018-05-23
fix: removed unnecessary cross-project dependencies

### LINKED ISSUE #325: Incorrect cross-package dependencies
Many of the packages depend on `webassemblyjs`, which results in them indirectly depending on everything else.

ColinEberhardt:
fixes #325 

I spotted this when adding the @webassembly/ast to astexplorer - it basically drags in everything via the dependency on `webassemblyjs`. This change removes these dependencies.

However ... a few projects depend on `webassemblyjs` just for `CompileError`. I've gone for the simple option, and created project-specific errors. I don't think reference equality of error constructors is that important?

### DIFF
--- a/packages/ast/package.json
+++ b/packages/ast/package.json
@@ -15,8 +15,7 @@
     "@webassemblyjs/helper-wasm-bytecode": "1.5.4",
     "@webassemblyjs/wast-parser": "1.5.4",
     "debug": "^3.1.0",
-    "mamacro": "^0.0.3",
-    "webassemblyjs": "1.5.4"
+    "mamacro": "^0.0.3"
   },
   "repository": {
     "type": "git",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,8 +21,7 @@
     "@webassemblyjs/ast": "1.5.4",
     "@webassemblyjs/wasm-parser": "1.5.4",
     "@webassemblyjs/wast-parser": "1.5.4",
-    "@webassemblyjs/wast-printer": "1.5.4",
-    "webassemblyjs": "1.5.4"
+    "@webassemblyjs/wast-printer": "1.5.4"
   },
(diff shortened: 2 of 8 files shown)


xtuc:
The issue is that the JS API has:
- WebAssembly.CompileError
- WebAssembly.LinkError
- WebAssembly.RuntimeError

So I think that we need to have them in the webassemblyjs package and be referenced by the other one.

ColinEberhardt:
> It's strange that no tests are failing. Are the error objects now duplicated?

Yes, I duplicated them, so wasm and wast parsers have their own `CompileError`. Perhaps not the best fix!

> So I think that we need to have them in the webassemblyjs package and be referenced by the other one.

The alternative is to have them in a `@webassembly\helper-errors` package - I'd be happy to make that change. I don't like all the unnecessary code we're dragging in to astexplorer, just for that on error!

(although I haven't checked whether astexplorer is tree shaking)

xtuc:
> Yes, I duplicated them, so wasm and wast parsers have their own CompileError. Perhaps not the best fix!

The issue is that we can't compare the errors anymore because they are two different objects (I think)?

> helper-errors

Yes sounds like a good solution to me. It seems to be a bigger issue we will have with the JS API (WebAssembly.Memory it used at many places).

I would like to regroup them using the same naming schema: `@webassemblyjs/helper-api-error`, `@webassemblyjs/helper-api-memory`, etc. What do you think?

I think that the `webassemblyjs` package should ideally be the dependence of none of our packages.

ColinEberhardt:
> I think that the webassemblyjs package should ideally be the dependence of none of our packages.

agreed - with the exception of the repl.

> `@webassemblyjs/helper-api-error`

sounds good to me - I'll update this PR using that approach.

ColinEberhardt:
I've implemented that here #348

PR #348, opened 2018-05-24
feat: added a new helper-api-error package

### LINKED ISSUE #325: Incorrect cross-package dependencies

ColinEberhardt:
replaces #344 
fixes #325


packages/wasm-parser/src/decoder.js
```diff
@@ -1,6 +1,6 @@
 // @flow
 
-import { CompileError } from "webassemblyjs/lib/errors";
+const { CompileError } = require("@webassemblyjs/helper-api-error");
```
xtuc: can you use ES6 import here?

packages/wast-parser/src/number-literals.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 const Long = require("long");
 const parseHexFloat = require("@webassemblyjs/floating-point-hex-parser");
-const { CompileError } = require("webassemblyjs/lib/errors");
+const { CompileError } = require("@webassemblyjs/helper-api-error");
```
xtuc: same, can you use ES6 imports please

packages/wasm-parser/src/decoder.js
ColinEberhardt: Oops - good point. Fixed that

ColinEberhardt:
> Cool, that's also a good first step to remove our cyclic deps
indeed - only a few left now.

PR #349, opened 2018-05-24
Visit union types

### LINKED ISSUE #334: Allow traverse to target UnionType
For example Instruction or Expression:

```js
traverse(ast, {
  Instruction() {},
  Expression() {}
});
```

And if we add a Node type to every node we could remove my `Node` hack used to traverse every node.

ColinEberhardt:
fixes #334


packages/ast/test/traverse.js
```diff
@@ -47,6 +47,19 @@ describe("AST traverse", () => {
     assert.isTrue(called, "Module visitor has not been called");
   });
 
+  it("should call the union type visitors", () => {
+    const node = t.stringLiteral("fish");
+    let called = false;
+
+    traverse(node, {
+      Expression() {
```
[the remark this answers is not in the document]
ColinEberhardt: That’s a good point, as node is no longer a special case this is redundant

packages/ast/src/traverse.js
```diff
@@ -143,6 +130,15 @@ export function traverseWithHooks(
         visitors[type](path);
         after(type, path);
       }
+
+      const unionTypes = unionTypesMap[type];
```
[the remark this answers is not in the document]
ColinEberhardt: Good idea - I’ll update to throw if a bide with an unknown type is found

ColinEberhardt:
traverse now throws an error if an unknown node type is encountered - and removed the duplicate test

packages/ast/test/traverse.js
```diff
@@ -47,6 +47,13 @@ describe("AST traverse", () => {
     assert.isTrue(called, "Module visitor has not been called");
   });
 
+  it("should throw if an unknown node type is encountered", () => {
+    const node = t.stringLiteral("fish");
+    node.type = "foo";
+
+    assert.throws(() => traverse(node, {}), "Unexpected node type foo");
```
xtuc: My initial comment was about:
```js
traverse(a, {
  NotANode() {}
})
```

Where we can check if `type` is in `unionTypesMap`.

xtuc: But it's nice to assert for internal errors too

ColinEberhardt: Ah, I was looking at the problem from the other direction! I've added that check too - although it did require a bit more structure:

https://github.com/xtuc/webassemblyjs/pull/349/commits/7d159dd1d9578478e26b3503be1f89cc2f4035ce#diff-a8223d97b02cf2fd47dbcdb8869579c8R997

PR #363, opened 2018-05-28
use ieee754 package

xtuc:


packages/wasm-parser/test/fixtures/literals/actual.wat
```diff
@@ -0,0 +1,12 @@
...
+  (func
+   (i32.const 1)
+   (i64.const 1)
+
+   (f32.const 1)
+   (f64.const 1)
+
+   (f64.const 0.1)
+  )
+  (;f32.const 0.1;)
```
xtuc: - wast: 0.1
- wasm: 0.10000000149011612

It seems that we could use a Float32Array but it's not support everywhere.

Issue #357, opened 2018-05-28
Upgrade flow


xtuc:
We upgraded flow recently and now we are seeing weird errors:
```
/home/travis/build/xtuc/webassemblyjs/packages/ast/src/signatures.js
  17:11  error  References:  flowtype-errors/show-errors
```

maurobringolf:
Does this include https://travis-ci.org/xtuc/webassemblyjs/jobs/383412920? These confused me but I havent tried to fix them yet

xtuc:
Yes exactly.

PR #369, opened 2018-05-29
Nan and Inf support in wasm-parser

maurobringolf:
I still need to fix the f64 case, there are some issues in the wast-parser.


xtuc:
Would you have the time to try to fix https://github.com/xtuc/webassemblyjs/issues/357? I'm seeing the errosrs in the CI. Otherwise LGTM

maurobringolf:
I'm looking into it after fixing f64. Theres some problems with wast-parser in the f64 case I believe.

maurobringolf:
Implemented and fixed the f64 cases (including `wast-parser`). Remaining errors are these flow reference errors from #357

xtuc:
Thanks

PR #372, opened 2018-05-30
Traverse refactor

### LINKED ISSUE #350: AST remove() function doesn't work for nodes with multiple child node properties
The current remove function relies on being able to remove a node given its parent:

https://github.com/xtuc/webassemblyjs/blob/master/packages/ast/src/traverse.js#L7

However, a number of nodes have multiple properties which can themselves contain child nodes, e.g. `ModuleMetadata`

ColinEberhardt:
I've been tidying up the AST traverse code, improving flow typing, adding a `parentKey` property.

fixes #350

Following this, I'm going to start expanding the node path API, adding methods for inserting sibling nodes etc ...


package.json
```diff
@@ -15,6 +15,7 @@
     "@babel/preset-flow": "7.0.0-beta.34",
     "@babel/template": "7.0.0-beta.34",
     "babel-eslint": "8.0.3",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
```
xtuc: `@babel/plugin-proposal-object-rest-spread` Should be 7.0.0-beta.34

packages/ast/src/traverse.js
```diff
@@ -44,58 +18,75 @@ function findParent(
...
+  }
 
-    const parentNode = parentPath.node;
-    removeNodeInBody(node, parentNode);
+  const parentNode = parentPath.node;
+  // $FlowIgnore: References?
+  const parentProperty = parentNode[parentKey];
+  if (Array.isArray(parentProperty)) {
+    // $FlowIgnore: References?
+    parentNode[parentKey] = parentProperty.filter(n => n !== node);
```
xtuc: The `parentKey` is really clever :+1:

ColinEberhardt:
@xtuc this should be good to go ... I've implemented `insertBefore`, `insertAfter` and `stop` - I've also re-implemented `remove` :-)

xtuc:
Cool, I'm looking into it right now. It seems that some tests are failing, likely due to a regression.

xtuc:
I use this technique to detect changes in `wasm-edit` made during the traversal: https://github.com/xtuc/webassemblyjs/blob/master/packages/wasm-edit/src/index.js#L60-L73

ColinEberhardt:
I'm struggling to see how the wasm-edit tests are failing, I can see that `replaceWIth` is performing the expected replacement, yet the test still fails. I've also added additional tests ... confused!

xtuc:
(I'm back on mobile)

I tried the old Object.assign based implementation of replace which makes the test passes, maybe there a subtle diff.

ColinEberhardt:
No worries, I’ll see if I can work it out :-)

ColinEberhardt:
found the difference, in my updated implementation the node that you supply to `replaceWith` is not 'visited'

packages/wasm-edit/src/index.js
```diff
@@ -58,18 +58,21 @@ export function editWithAST(
...
   function after(type: string, path: NodePath<*>) {
     if (path.node._deleted === true) {
+      console.log(type, "deleted");
       operations.push({
         kind: "delete",
         node: path.node
       });
       // $FlowIgnore
     } else if (hashNode(nodeBefore) !== hashNode(path.node)) {
+      console.log(type, "different hash");
```
xtuc: btw please let me know if you have a better idea here? I also tried the tree diffing approach but was quite too complicated.

PR #387, opened 2018-06-06
fix: wasm parser adds generated identifiers to globals

ColinEberhardt:
Continuing to expand the wasm-parser tests when I have a few moments spare ...


xtuc:
Actually i'm trying to remove them.

This

``wat
(global i32 (i32.const 1))
``` 

Will be transformed to (in the AST):

``wat
(global $global_0 i32 (i32.const 1))
``` 

Which doesn't make sense anymore to me. That will cause issues in the `wasm-parser` where instead of having the index we will have an identifier to resolve.

PR #388, opened 2018-06-06
Proper dependency declarations

onlywei:
These improper dependency declarations cause these packages to fail when using a more strict client such as [pnpm](https://pnpm.js.org/).


xtuc:
Thanks for your PR.  


I feel like that happens all the time. Do you know if there a way to add it to our CI?

xtuc:
@onlywei since we are using a lerna to install the dependencies in the CI, i'm not sure what we could do.

onlywei:
@xtuc Thanks. Let me know when a release is cut. Webpack is broken without this.

PR #389, opened 2018-06-07
feat: add test pnpm

xtuc:


xtuc:
FYI @onlywei, i'm trying to add a check in our CI. Maybe you have thoughts on that?

Issue #355, opened 2018-05-28
initializer expression cannot reference a mutable global


xtuc:
Found in https://github.com/webpack/webpack/pull/7404

maurobringolf:
Would this fall into the `isConst` check? i.e.

https://github.com/xtuc/webassemblyjs/blob/master/packages/validation/src/is-const.js#L26-L28

should check the moduleContext for mutability of the global somehow?

xtuc:
Yes that's correct. The issue is that this check is missing too much context. We should refactor it and provide the `ModuleContext`.

PR #392, opened 2018-06-07
Refactor validation

maurobringolf:
Some refactor of validations improving a couple of things:

* Mutable globals cannot be used in global initialization Fixes https://github.com/xtuc/webassemblyjs/issues/355.

* ModuleContext is created before running validations and passed to the individual steps.

* Validation `fixtures` now run all validation steps, not just type-checking and I removed the separate `isConst` tests since its API changed to `Program -> Array<string>` giving a list of errors too.

We should probably merge all validation visitors and then run one big `traverse` doing all the work at the same time?


packages/validation/src/index.js
```diff
@@ -16,7 +15,21 @@ export default function validateAST(ast: Program) {
...
+export function getValidationErrors(ast: Program): Array<string> {
+  const errors = [];
+
+  ast.body.filter(({ type }) => type === "Module").forEach(m => {
+    const moduleContext = moduleContextFromModuleAST(m);
+
+    errors.push(...isConst(ast, moduleContext));
+    errors.push(...importOrderValidate(ast));
+    errors.push(...typeChecker(ast, moduleContext));
+  });
```
maurobringolf: There are some inconsistencies in the way the `repl` constructs modules. It doesnt always pass a `Program` here, but sometimes `ast` is directly a `Module` node itself. I tried checking for it here, but it didnt fix all the tests, so it probably requires some changes to the repl?

maurobringolf: I found the problem. The repl also used the type checker directly. I switched to the new  `getValidationErrors` API.

maurobringolf:
@xtuc I finally got around fixing this, except for one spec test. But I am not sure how it relates to the current changes in the specification: https://github.com/WebAssembly/spec/pull/814. Do you know when the official testsuite is updated?

xtuc:
@maurobringolf (sorry for the delai). I tried to upgrade our spec tests here https://github.com/xtuc/webassemblyjs/pull/296, I will try to finish it eventually. 

The issue with mutable global is that you can't really use them atm, no support but it's coming soon

PR #404, opened 2018-06-30
Fixing cyclic dependency

sendilkumarn:
r? @xtuc


packages/helper-module-context/test/index.js
```diff
@@ -1,30 +1,25 @@
...
     it("should return the start function offset", () => {
-      const context = contextFromWast(`
-        (module
-          (func)
-          (func)
-          (start 1)
-        )
-      `);
+      const context = contextFromWast(
+        '{"type":"Module","id":null,"fields":[{"type":"Func","name":{"type":"Identifier","value":"func_0","raw":""},"signature":{"type":"Signature","params":[],"results":[]},"body":[],"loc":{"start":{"line":3,"column":9},"end":{"line":3,"column":12}}},{"type":"Func","name":{"type":"Identifier","value":"func_1","raw":""},"signature":{"type":"Signature","params":[],"results":[]},"body":[],"loc":{"start":{"line":4,"column":9},"end":{"line":4,"column":12}}},{"type":"Start","index":{"type":"NumberLiteral","value":1,"raw":"1"},"loc":{"start":{"line":5,"column":9},"end":{"line":5,"column":15}}}],"loc":{"start":{"line":2,"column":7},"end":{"line":6,"column":5}}}'
```
xtuc: I think it was more readable with the Wast format. I'm wondering why you needed to change it?

xtuc: Ah for the devdep? That's fine

packages/node-helper/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@webassemblyjs/node-helper",
+    "version": "1.5.13",
+    "description": "WebAssembly Node Helper",
+    "main": "lib/index.js",
+    "scripts": {
+      "test": "echo \"Error: no test specified\" && exit 1"
+    },
+    "author": "Sendil Kumar N",
```
xtuc: The issue is that most of the code in node-helper was actually from other contributors and me. Is that correct?

packages/wasm-parser/test/index.js
```diff
@@ -66,6 +65,11 @@ describe("Binary decoder", () => {
 
   compare(testSuites, getActual, getExpected);
 
+  function makeBuffer(...splitedBytes) {
+    const bytes = [].concat.apply([], splitedBytes);
+    return new Uint8Array(bytes).buffer;
+  }
```
xtuc: Dev deps are fine we can keep the helper here or there's another reason?

sendilkumarn: Lerna will complain about it even when it is devdep

packages/node-helper/package.json
sendilkumarn: ha ha... `yarn init --yes` 😉 I will change it

packages/wast-printer/test/index.js
```diff
@@ -4,14 +4,13 @@ const {
...
 } = require("@webassemblyjs/helper-test-framework");
-const { parse } = require("@webassemblyjs/wast-parser");
 
 const { print } = require("../lib/index");
 
 describe("printer", () => {
   describe("wast", () => {
-    const testSuites = getFixtures(__dirname, "fixtures", "**/actual.wast");
-    const pre = f => print(parse(f));
+    const testSuites = getFixtures(__dirname, "fixtures", "**/actual.json");
```
xtuc: I'm not sure why we need to change that?

packages/node-helper/src/index.js
```diff
@@ -0,0 +1,58 @@
+// @flow
```
xtuc: Is this intentional? This should probably stay in the AST package.

packages/node-helper/src/nodes.js
```diff
@@ -0,0 +1,1044 @@
+// @flow
```
xtuc: Is this intentional?

packages/node-helper/src/index.js
```diff
@@ -0,0 +1,58 @@
...
+  parse32F,
+  parse64F,
+  parse32I,
+  parse64I,
+  parseU32,
+  isNanLiteral,
+  isInfLiteral
+} = require("./number-literals");
+
+import { longNumberLiteral, floatLiteral, numberLiteral } from "./nodes";
```
xtuc: Now I understand why you need https://github.com/xtuc/webassemblyjs/pull/404/files#r200809128.

Duplicated `nodes.js` to avoid the cyclic dep is not ideal IMO

sendilkumarn: Ahaha.. this is `auto-generated` right? I think some refactoring needed.

sendilkumarn: actually it depends on it. I think I have to redirect things here. But it is everywhere. I will find sometime this week

xtuc: ok thanks!

PR #408, opened 2018-07-07
Add node type assertions

xtuc:


packages/ast/src/definitions.js
```diff
@@ -394,6 +394,7 @@ defineType("Table", {
       type: "TableElementType"
     },
     limits: {
+      assertNodeType: true,
```
xtuc: I'm introducing this since it's easier for AST breaking changes, we could make it default at some point (doesn't handle the union correctly atm)

Issue #405, opened 2018-07-03
wasm-edit -- Valid edits can throw with default decode options.


TimHambourger:
### Description
Because of the `assertNotIdentifier` calls in [wasm-gen/lib/encoder/index.js](https://github.com/xtuc/webassemblyjs/blob/master/packages/wasm-gen/src/encoder/index.js), editing, say, a module export name can throw with the default decode options.

### Steps to repro
Start with input WASM corresponding to the WAT
```wat
(module
  (func $add (param $lhs i32) (param $rhs i32) (result i32)
    get_local $lhs
    get_local $rhs
    i32.add)
  (export "add" (func $add))
)
```
(from [MDN's examples](https://github.com/mdn/webassembly-examples/blob/master/understanding-text-format/add.wat)).

Then edit the export name like
```js
import { edit } from '@webassemblyjs/wasm-edit';

const bin; // ArrayBuffer containing .wasm contents

edit(bin, {
    ModuleExport(path) {
        path.node.name += '!!';
    }
});
```

This throws with "Unsupported node Identifier" due to the assertion at
```js
function encodeModuleExport(n) {
  var out = [];
  assertNotIdentifierNode(n.descr.id);
  // etc.
}
```

If you instead decode with `ignoreCustomNameSection: true` then the edit succeeds:
```js
import { decode } from '@webassemblyjs/wasm-parser';
import { editWithAST } from '@webassemblyjs/wasm-edit';

const bin; // ArrayBuffer containing .wasm contents

const ast = decode(bin, {
    ignoreCustomNameSection: true
});

editWithAST(ast, bin, {
    ModuleExport(path) {
        path.node.name += '!!';
    }
});
```
At the relevant assertion, `n.node.descr` now has type `'NumberLiteral'` instead of `'Identifier'`.

### Expected behavior
That valid edits succeed with default decode options. The current behavior feels restrictive.

### Other Cases?
This behavior with export names seems to repro reliably across input .wasm files. I also found a possible repro with import names, but that one only affected some of the files I tested and might be due to a bug with `decode`. I'll file that separately.

ColinEberhardt:
Thanks for the detailed report. As you've no doubt worked out this issue is due to the presence of a [custom name section](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#name-section) within your wasm binary.

[...]

PR #409, opened 2018-07-07
refactor index

xtuc:
Attempt to fix https://github.com/xtuc/webassemblyjs/issues/405

The idea is to transform our `Index` that has either `Identifier` or `NumerLiteral` to a structure that contains both (the Identifier is optional tho).

Notes:
- The naming is awful (`indexOfIndex`), any idea?

### DIFF
--- a/packages/ast/src/definitions.js
+++ b/packages/ast/src/definitions.js
@@ -272,7 +272,8 @@ defineType("Elem", {
   unionType: ["Node"],
   fields: {
     table: {
-      type: "Index"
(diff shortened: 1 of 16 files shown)


ColinEberhardt:
For the naming, it might be worth using the terminology from the specification:

image

They refer to identifiers as being either numeric or symbolic.

This suggests to me something like the following:

```
{
   type: "identifier",
   symbolic: "log",
   numeric: 23
}
```

PR #410, opened 2018-07-12
wasm-gen -- Support encoding vectors with multi-byte encoded lengths

TimHambourger:
Do proper LEB encoding from `encodeVec`. This enables generating strings longer than 127 bytes, like long import/export names.

### DIFF
--- a/packages/wasm-gen/test/index.js
+++ b/packages/wasm-gen/test/index.js
@@ -238,18 +238,22 @@ const fixtures = [
 
   {
     name: "0123456789 (x13)",
-    node: (() => {
-      const unit = "0123456789";
-      let str = "";
-      for (let i = 0; i < 13; i++) str += unit;
-      return t.stringLiteral(str);
-    })(),
-    expected: (() => {
-      const unit = [0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39];
-      let expected = [0x82, 0x01];
-      for (let i = 0; i < 13; i++) expected = expected.concat(unit);
-      return expected;
-    })()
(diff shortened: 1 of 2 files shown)


packages/wasm-gen/test/index.js
```diff
@@ -234,6 +234,22 @@ const fixtures = [
     name: "./-Ɂ?_¶",
     node: t.stringLiteral("./-Ɂ?_¶"),
     expected: [9, 0x2e, 0x2f, 0x2d, 0xc9, 0x81, 0x3f, 0x5f, 0xc2, 0xb6]
+  },
+
+  {
+    name: "0123456789 (x13)",
+    node: (() => {
+      const unit = "0123456789";
```
xtuc: You could use `"0123456789".repeat(13)` here (available since Node 0.12)

packages/wasm-gen/test/index.js
```diff
@@ -234,6 +234,22 @@ const fixtures = [
...
+      let str = "";
+      for (let i = 0; i < 13; i++) str += unit;
+      return t.stringLiteral(str);
+    })(),
+    expected: (() => {
+      const unit = [0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37, 0x38, 0x39];
+      let expected = [0x82, 0x01];
+      for (let i = 0; i < 13; i++) expected = expected.concat(unit);
```
xtuc: You could maybe use `Array(13).fill(..)` but that's not very convenient to use, so it's up to you.

TimHambourger:
@xtuc. Thanks for the suggestions! I'll update

TimHambourger:
Thanks!

PR #411, opened 2018-07-13
fix: module imports should not use type index as their function identifier

### LINKED ISSUE #406: wasm-parser -- Decode can produce unexpected module import identifier
### Description
In certain cases, `decode` can produce AST containing unexpected identifiers inside module import nodes.

### Steps to Repro
The behavior depends highly on the input .wasm file. I haven't found a good minimal repro yet, but you can use [this file](https://github.com/TimHambourger/webassemblyjs-wasm-corruption/blob/master/webassemblyjs-repro/wasm-input/get37_release_bg.wasm) as input. Note that that file was compiled in release mode from Rust src that uses [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen), then post-processed with the wasm-bindgen CLI. The Rust src is [quite trivial](https://github.com/TimHambourger/webassemblyjs-wasm-corruption/blob/master/wasm-src/get37/src/lib.rs).

With that input in hand, decode like
```js
import { decode } from '@webassemblyjs/wasm-parser';

const bin; // ArrayBuffer containing the .wasm input

const ast = decode(bin);

// Note that there's only one import in the file
const importNode = ast.body[0].fields.filter(f => f.type === 'ModuleImport')[0];
```
At this point, importNode looks like
```json
{
  "type": "ModuleImport",
  "module": "./get37",
  "name": "__wbindgen_throw",
  "descr": {
    "type": "FuncImportDescr",
    "id": {
      "type": "Identifier",
      "value": "__rust_dealloc"
    },
    "signature": {
      "type": "Signature",
      "params": [
        {
          "valtype": "i32"
        },
        {
          "valtype": "i32"
        }
      ],
      "results": []
    }
  },
  "loc": {
    "start": {
      "line": -1,
      "column": 120
    },
    "end": {
      "line": -1,
      "column": 147
    }
  }
}
```

`module` and `name` are as expected, but `descr.id` is unexpected. If you process the same input w/ [wasm2wat](https://github.com/WebAssembly/wabt), the relevant excerpts look like
```wat
(module
  (type $t5 (func (param i32 i32)))
  (import "./get37" "__wbindgen_throw" (func $__wbindgen_throw (type $t5)))
  (func $core::ptr::drop_in_place::h99e3afcc3d3d20ab (type $t0) (param $p0 i32))
  (func $__wbg_get37_free (export "__wbg_get37_free") (type $t0) (param $p0 i32)
    (drop
      (call $wasm_bindgen::convert::GlobalStack::new::hf5174142e84b82ff))
    (block $B0
      (block $B1
        (br_if $B1
          (i32.eqz
            (get_local $p0)))
        (br_if $B0
          (i32.load
            (get_local $p0)))
        (i32.store align=1
          (get_local $p0)
          (i32.const 0))
        (call $__rust_dealloc
          (get_local $p0)
          (i32.const 4)
          (i32.const 4))
        (return))
```
So you can see the `import` node and some time later the `__rust_dealloc` identifier, but unless I'm missing something `__rust_dealloc` is NOT the identifier for the import node's description.

### Expected behavior
That `importNode.descr.id.value` would be `'__wbindgen_throw'`. Unless I'm missing something.

### Remarks
Downstream, this can lead to exceptions editing import names as mentioned in #405. That repro goes like:
```js
import { edit } from '@webassemblyjs/wasm-edit';

const bin; // ArrayBuffer containing .wasm input
edit(bin, {
    ModuleImport(path) {
        path.node.name += '!!';
    }
});
```

That throws with "Unsupported node Identifier" due to the assertion in
```js
    case "FuncImportDescr":
      {
        out.push(0x00); // $FlowIgnore

        assertNotIdentifierNode(n.descr.id); // $FlowIgnore

        out.push.apply(out, _toConsumableArray(encodeU32(n.descr.id.value)));
        break;
      }
```

At that assertion, `n.descr.id.value` is `'__rust_dealloc'`.

As in #405, the exception can be avoided by `decode`ing with `ignoreCustomNameSection: true`. At that point, the output WASM seems to be what I'd expect, based on running back through wasm2wat.

ColinEberhardt:
fixes #406 

@xtuc there is a wasm-edit test failing as a result of this change, wondering if you have any pointers?


packages/wasm-parser/test/fixtures/import/complex/actual.wat
```diff
@@ -0,0 +1,5 @@
+(module
```
ColinEberhardt: this simple test, which adds a couple of type definitions, was enough to repro the issues from #406

packages/wast-parser/src/grammar.js
```diff
@@ -419,6 +417,8 @@ export function parse(tokensList: Array<Object>, source: string): Program {
         const fnParams = [];
         const fnResult = [];
 
+        let fnName = t.identifier(getUniqueName("func"));
```
ColinEberhardt: this change was not necessarily needed to make tests pass, but this does mean we have the same generated names when parsing wat

xtuc:
The "AST synchronization" test suite in `wasm-edit` ensure that we mutate the AST correctly by comparing it to a freshly decoded AST.

packages/wasm-parser/src/decoder.js
```diff
@@ -432,7 +432,7 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
           throw new CompileError(`function signature not found (${typeindex})`);
         }
 
-        const id = t.numberLiteralFromRaw(typeindex);
+        const id = getUniqueName("func");
```
xtuc: Maybe we should try to eliminate the unique name generator (because statefull), and rather use the `typeindex` here.

packages/wast-parser/src/grammar.js
xtuc: Following https://github.com/xtuc/webassemblyjs/pull/411/files#r202406723 we could add our own func increment, or maybe we already have this information?

ColinEberhardt:
> The "AST synchronization" test suite in wasm-edit ensure that we mutate the AST correctly by comparing it to a freshly decoded AST.

@xtuc ok, I've had a look at the test and understand what it is doing now. I've fixed the test by making the following changes:

- the `makeFuncNodes` function was adding the AST for the function type as well as the function itself. I've moved the construction of function types into a separate step as the module imports use this too
- I've changed the order, to add the module import first
- following this, the test adds a function, which now has an index of 1, and an identifier of `func_1`. As discussed in https://github.com/xtuc/webassemblyjs/pull/409 - this is something that we'll probably remove in future, but for now, this fixes the original issue where type indices were being used as function indices.

ColinEberhardt:
The tests are currently failing, but looks like a CI issue rather than a unit test failure!

ColinEberhardt:
kicked Travis and the build works now :-)

PR #417, opened 2018-07-18
Ignore custom section decoding errors

xtuc:
Spec: https://webassembly.github.io/spec/core/binary/modules.html#custom-section
> If an implementation interprets the contents of a custom section, then errors in that contents, or the placement of the section, must not invalidate the module.

@TimHambourger's test suite passes now, but it shows a lot of custom section decoding errors (`invalid UTF-8 encoding`).

The spec only mentions the content of the section, we need to figure out why the name is failing.


xtuc:
Nervermind, that's wrong. I should rather fix the root of the issue.

Issue #414, opened 2018-07-14
Wasm-parser gives "invalid UTF-8 encoding" on intermediate output from wasm-bindgen


TimHambourger:
I'm filing this to track the remaining [failing cases](https://github.com/xtuc/webassemblyjs/issues/407#issuecomment-404995901) from #407 that aren't addressed by #413. From now on I'll consider these separate from #407.

Again, those cases are
```
get37_debug.wasm -- addAssignExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_debug.wasm -- oneCharExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_debug.wasm -- removeExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_debug.wasm -- renameImports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_debug.wasm -- addAssignExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_debug.wasm -- oneCharExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_debug.wasm -- removeExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_debug.wasm -- renameImports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_release.wasm -- addAssignExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_release.wasm -- oneCharExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_release.wasm -- removeExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_no_struct_release.wasm -- renameImports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
    native WebAssembly error: null
get37_release.wasm -- addAssignExports -- FAIL
    wasm-parser error:        Error: invalid UTF-8 encoding
[...]

PR #418, opened 2018-07-18
Fix decoding types name section

xtuc:
Basically:
- `name_payload_len` field was missing in module name (don't think we noticed it, we can't even test it because of missing feature upstream).
- `varuint7` instead of readByte (we need to use LEB128 to decode it (0xxxxxxx) otherwise the first bit would end up in the actual value)
- `varuint32` (doesn't really matter since it's variable, but now it's limited to 4 bytes)

cc @ColinEberhardt since it was quite interesting.

Ref https://github.com/xtuc/webassemblyjs/issues/414

### DIFF
--- a/packages/leb128/src/index.js
+++ b/packages/leb128/src/index.js
@@ -3,13 +3,13 @@
 import leb from "./leb";
 
 /**
- * According to https://webassembly.github.io/spec/binary/values.html#binary-int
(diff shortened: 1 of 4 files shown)


ColinEberhardt:
Great - thanks for the heads-up, I should have read the specification more closely https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#name-section

PR #420, opened 2018-07-18
Add ESM build

xtuc:


xtuc:
@mathiasbynens could you please try with the `v1.7.1-0` which has the `module` key + an ESM build (on each package).

Unfortunately, it doesn't seem to work on bundle.run (https://bundle.run/@webassemblyjs/wasm-parser@1.7.1-0 or https://bundle.run/@webassemblyjs/wast-printer@1.7.1-0/esm/index.js) which uses Rollup under the hood.

packages/ast/src/utils.js
```diff
@@ -1,8 +1,8 @@
 // @flow
 import { signatures } from "./signatures";
 import { traverse } from "./traverse";
-import constants from "@webassemblyjs/helper-wasm-bytecode";
 
+const constants = require("@webassemblyjs/helper-wasm-bytecode");
```
mathiasbynens: I'm confused. We're trying to make it possible for folks to consume these utilities as modules, right? This change seems to go against that. Why is this needed?

xtuc: I'm taking advantage of the CJS interop because I don't have an ESM default export and rollup wasn't able to resolve it.

mathiasbynens:
With `@webassemblyjs/wast-printer@1.7.1-0`, and this code:

```js
import { print } from '@webassemblyjs/wast-printer';

print(/* ... */);
```

I get the following Rollup warning:

```
https://github.com/rollup/rollup/wiki/Troubleshooting#treating-module-as-external-dependency
tty (imported by node_modules/debug/src/node.js, commonjs-external:tty)
util (imported by node_modules/debug/src/node.js, commonjs-external:util)
os (imported by node_modules/supports-color/index.js, commonjs-external:os)
created public/bundle.mjs in 994ms
```

Looks like `wast-printer`, through one of its dependencies, pulls in some Node-specific code.

This seems unnecessary in this case; wast-printer only needs the input WASM AST and then to return a string.

xtuc:
The module indirectly depends on `debug` which seems to depends on os/tty etc. I'll simply remove that.

packages/ast/src/utils.js
xtuc: Ok I understand now that was wrong, i'm working on the browser compat of `wast-printer`.

xtuc:
@mathiasbynens `wast-printer@1.7.0-3` works now (note that we have a circular dep).

If you can confirm that it works on your side, i'll merge this.

mathiasbynens:
Progress!

I’m now getting an error on this line in `leb.js`, because `require` doesn’t exist:

```
var bits = require("./bits");
```

xtuc:
Ah, are you bundling `wasm-parser`? I successfully bundled `wast-printer` but that might be a bug on my side.

mathiasbynens:
Ah, sorry, yeah I’m trying to bundle both `wasm-parser` and `wast-printer`.

mathiasbynens:
`wast-printer` already works indeed! Thank you!

xtuc:
I'd like to merge that first (before it becomes unreviewable) and I'll work on wasm-parser then.

mathiasbynens:
Sounds good to me!

PR #422, opened 2018-07-19
wasm-parser use esm

xtuc:


xtuc:
@mathiasbynens `@webassemblyjs/wasm-parser` is now ESM compatible too. https://bundle.run/@webassemblyjs/wasm-parser@1.7.1 works!

xtuc:
Unfortunately, I just noticed that we still rely on Node's Buffer. I'll have to pull in a polyfill for now, ideally we switch to ArrayBuffer.

xtuc:
I just released `1.7.2` with a polyfill.

xtuc:
@mathiasbynens thinking about it more, including the polyfill was a bad idea. Some peoples are consuming the project as a library and don't need the polyfill.

Is there a way you can add it into your bundling? You can use a module alias.

PR #424, opened 2018-07-19
remove Node buffer (web compat)

xtuc:


xtuc:
It has too much impact, let's polyfill it instead.

PR #427, opened 2018-07-20
Correct byte counting when parsing custom sections

### LINKED ISSUE #414: Wasm-parser gives "invalid UTF-8 encoding" on intermediate output from wasm-bindgen

TimHambourger:
Fixes #414.


packages/wasm-parser/src/decoder.js
```diff
@@ -1752,12 +1752,17 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
           t.sectionMetadata("custom", startOffset, sectionSizeInBytesNode)
         ];
 
+        const offsetBeforeReadUTF8 = offset;
         const sectionName = readUTF8String();
+        const bytesEatenByReadUTF8 = offset - offsetBeforeReadUTF8;
         eatBytes(sectionName.nextIndex);
```
xtuc: `sectionName.nextIndex` already contains the `bytesEatenByReadUTF8` information, it's the amount of bytes that we can eat after the read.

packages/wasm-parser/src/decoder.js
```diff
@@ -1752,12 +1752,17 @@ export function decode(ab: ArrayBuffer, opts: DecoderOpts): Program {
...
 
         dump([], `section name (${sectionName.value})`);
 
-        const remainingBytes = sectionSizeInBytes - sectionName.nextIndex;
+        // B/c readUTF8String includes an eatBytes call, we have to account
+        // for bytes eaten by readUTF8String itself
```
xtuc: That's a good catch! My previous comment (that you can ignore then) assumed that. That's confusing.

packages/wasm-parser/test/custom-sections.js
```diff
@@ -0,0 +1,556 @@
...
+              }
+            }
+          ]
+        }
+      }
+    ]
+  };
+}
+
+function expectedForCustomAfterName() {
```
xtuc: Could we use a JSON string for the AST instead? You can then compare it (line by line with good errors) using https://github.com/xtuc/webassemblyjs/blob/master/packages/helper-test-framework/src/index.js#L35

xtuc:
I have a suggestion, `readUTF8String` doesn't return the correct `nextIndex` and I guess it impacts some other cases.

If you look at the code here https://github.com/xtuc/webassemblyjs/blob/master/packages/wasm-parser/src/decoder.js#L195-L211, what if we set `nextIndex = strlen + lenu32.nextIndex` where `strlen` is the length and `lenu32` is the vector size.

TimHambourger:
> what if we set nextIndex = strlen + lenu32.nextIndex where strlen is the length and lenu32 is the vector size

Yes, I like that. But on it's own that'll break expected usages like
```js
const name = readUTF8String();
eatBytes(name.nextIndex);
```
since some of those bytes will have already been eaten by `readUTF8String` itself.

So another way to say it is that currently `readUTF8String` has a net effect on the current `offset`. Which sets it apart from all other `readXXX` functions, which are neutral with respect to the current `offset`.

So what if we make `readUTF8String` neutral w.r.t. the `offset`? We can't just ditch the `eatBytes` call inside `readUTF8String`, since it's setting up the offset for the following `readBytes` call. But we can reset the `offset` afterwards:
```js
  function readUTF8String(): DecodedUTF8String {
    const lenu32 = readU32();
    // Setup offset for the readBytes call below
    eatBytes(lenu32.nextIndex);

    const strlen = lenu32.value;

    dump([strlen], "string length");

    const bytes = readBytes(strlen);

    // Reset offset so that readUTF8String is neutral w.r.t. the offset
    // It's up to consuming code to call eatBytes afterwards
    eatBytes(-lenu32.nextIndex);

    const value = utf8.decode(bytes);

    return {
      value,
      nextIndex: strlen + lenu32.nextIndex
    };
}
```

Thoughts?

packages/wasm-parser/test/custom-sections.js
xtuc: Ok I see, in that case I would recommend you using `traverse` and target the nodes directly.

TimHambourger:
@xtuc. Yes, I like your `readBytesAtOffset` suggestion. I'll go that route and update this PR shortly.

xtuc:
Btw CI seems fine, `test-pnpm` should be optional I guess and the first one is a timeout.

Issue #394, opened 2018-06-09
Refac stack unwindings


xtuc:
see https://github.com/WebAssembly/wabt/issues/854#issuecomment-394533477

PR #449, opened 2018-08-26
Proper compiler and IR

xtuc:
Fixes https://github.com/xtuc/webassemblyjs/issues/394, and a couple of related issues.


packages/helper-compiler/src/index.js
```diff
@@ -0,0 +1,46 @@
...
+
+import { Module } from "./module";
+
+export function toIR(ast: Program): IR {
+  const program = {};
+  const funcTable = [];
+
+  // flatten the ast first
+  // TODO(sven): do it in a single pass?
+  flatten(ast);
```
xtuc: The compability with wast might be broken here anyway, we can remove the AST flatten pass.

xtuc:
Ok i'm in a state were all the regression were fixed

xtuc:
Ok we're in mergable state here. I need to refac a bit.

PR #455, opened 2018-09-19
fix: Upgrade lerna to 3, fixes #452

### LINKED ISSUE #452: Add LICENSE files to packages/*
Hello,

It looks like packages/floating-hex-parser has a LICENSE file, but most of the other packages don't. Would you mind adding LICENSE files to the other packages, too?

pksunkara:
This PR upgrades lerna from version 2 to version 3. It also makes use of yarn workspaces.

1. Since `useWorkspaces: true` is present in `lerna.json`, doing `yarn install` will also bootstrap lerna. This also allows us to define the packages in only `package.json`
2. lerna@3 will always use the local installed version which means we can remove lerna version from `lerna.json`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: node_js
 
-before_script: make bootstrap
+cache:
+  yarn: true
+
+before_install: npm i -g yarn
+install: make bootstrap
 
 script: make $JOB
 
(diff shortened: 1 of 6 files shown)


pksunkara:
We would like to ideally get the new releases with LICENSE in the upcoming webpack patch release. Just waiting for you 😄

PR #457, opened 2018-09-19
Fix AppVeyor build by installing correct version of Lerna

kuba-kubula:


pksunkara:
Fixed this in #455. This is a temporary solution until that is merged.

PR #459, opened 2018-09-20
Better readability for `make bootstrap`

pksunkara:
Basically, `lerna boostrap` with the configuration you have is the same as `yarn install`. But I made this change for better understanding of the Makefile for future contributors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ clean-all:
 	rm -rf ./node_modules ./packages/*/node_modules ./packages/*/{lib,esm}
 
 bootstrap: clean-all
-	yarn install
+	$(LERNA) bootstrap
 
 build:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ clean-all:
 	rm -rf ./node_modules ./packages/*/node_modules ./packages/*/{lib,esm}
 
 bootstrap: clean-all
-	yarn install
+	$(LERNA) bootstrap
```
xtuc: The CI is failing because `lerna` is not installed, I'm using the local executable provided during the `yarn install`. Either we install lerna manually before or leave it as is.

pksunkara: You are right. Closing this.

PR #460, opened 2018-09-26
Make compilier compatible with WAST

xtuc:


xtuc:
I tried to make the compiler compatible with wast because of the spectest runner, but it seems that we can workarround that.

PR #464, opened 2018-10-18
chore: fix sort to work on new node version

sendilkumarn:
This fixes the sorting problem w.r.t node-nightly builds

r? @xtuc


packages/wasm-edit/src/index.js
```diff
@@ -31,7 +31,7 @@ function sortBySectionOrder(nodes: Array<Node>) {
     }
 
     // $FlowIgnore ensured above
```
sokra: Can the FlowIgnore can be removed now? Did it hide the error?

xtuc: Yes indeed, thanks.

PR #484, opened 2019-02-13
feat: enable tests

xtuc:


xtuc:
Just need to remove from-wasm-bindgen tests, which shows a wrong error (probably caused by try/catching a lot)

PR #489, opened 2019-02-17
fix: data printing of non-printable characters

menduz:
Some characters were printed using JSON notation.

Previously a data with `[0, 6]` were printed as
```lisp
(data 0 (i32.const 2) "\u0000\u0006")
```
Now it prints
```lisp
(data 0 (i32.const 2) "\00\06")
```

The condition of printable characters was taken from [WABT](https://github.com/WebAssembly/wabt/blob/master/src/wat-writer.cc#L47).

It now fails printing invalid bytes (>255)


xtuc:
Thanks for your PR!

In Wabt `[0, 6]` prints as `[0, 6]`, unfortunately, the Wast "spec" doesn't document the `chars`. Let's maybe align on Wabt?

The wast-parser should fail when special UTF8 chars are found (that's how wabt does it), could you please test it too?

Also, I would like to try printing with UTF8 chars, but we'll need to construct an AST ourselves. I'm happy to commit this test case if needed.

menduz:
I took the UTF8 information from [data segments](http://webassembly.github.io/spec/core/text/modules.html#text-datastring) which defines [strings here](http://webassembly.github.io/spec/core/text/values.html#text-string) as UTF8.

I updated the PR with a [WABT compatible condition instead of JSON](https://github.com/WebAssembly/wabt/blob/master/src/wat-writer.cc#L47).

I also added a couple of UTF8 examples, please validate in your side. This is ready to merge, already tested it with a real world scenario and it is compatible back and forth with binaryen and wabt.

xtuc:
I published the version 1.8.4 (of all the packages).

menduz:
TY

PR #695, opened 2019-11-30
move mamacro to dev-dependencies in helper-compiler and helper-module…

### LINKED ISSUE #578: mamacro should be a devDependency for helper-module-context
It looks like `mamacro` is listed as a dependency, instead of a devDependency, in https://github.com/xtuc/webassemblyjs/blob/573c38515f80d582ebebc357ace5742197450551/packages/helper-module-context/package.json

My understanding from https://github.com/xtuc/mamacro/issues/2 is that `mamacro` should only be listed as a dev dependency.

The reason why this is a problem for us is that `mamacro` doesn't have a LICENSE file, but since it's listed as a dependency instead of a devDependency, it shows up in our license scanner.

Thanks :)

momesana:
mamacro should be a devDepenceny as outlined by @xtuc in https://github.com/xtuc/mamacro/issues/2.

Resolves #578

### DIFF
--- a/packages/helper-compiler/package.json
+++ b/packages/helper-compiler/package.json
@@ -13,12 +13,12 @@
   },
   "dependencies": {
     "@webassemblyjs/ast": "1.8.5",
-    "@webassemblyjs/helper-flatten-ast": "1.8.5",
-    "mamacro": "^0.0.7"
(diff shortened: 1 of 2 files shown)


momesana:
Thanks a lot. Hope it will be merged soon :-).

PR #968, opened 2021-08-30
Implement offset memarg in wasm-parser and wast-printer and close #956

### LINKED ISSUE #956: wasm/wast-parser inconsistency: "Mem" vs "Memory"
The statement `(export "memory" (memory 0))` gets parsed by wast-parser to an AST node with `node.descr.type` equal to `"Memory"`. On the other hand, wasm-parser produces `node.descr.type` equal to `"Mem"`.

I thought I'd open an issue because I'm not sure such inconsistencies are wanted.

mitschabaude:
@xtuc Please review!

This PR contains three things related to memory / memargs:

* First commit: As discussed in #956, change the `node.descr.type` of memory exports/imports from "mem" to "memory" in wasm-parser, to be consistent with what wast-parser produces

* Second commit: Implement named instruction arguments (e.g. the offset `i32.load offset=4`) in wasm-parser to support parsing offsets in memory load/store instructions. These named arguments were already supported by wast-parser and I made sure that wasm-parser now produces the same AST, by adding a property `namedArgs` to instruction nodes which is an object, e.g. `{offset: numberLiteral(...)}`.

Two tests were failing after this change, which actually are tests that supposedly tested for offset support (which didn't exist). It turned out that the tests were wrong and it seemed that they were deliberately re-written to not fail *because* of missing offset support. The tests are fixed now and succeed with the new implementation, so this new feature is already covered by the test suite :)

* Third commit: Implement the printing of named arguments in wast-printer, which means wast-printer also supports the offset memarg now.

These changes might need a review because I am not familiar with flow syntax yet and might have added code that does not conform to the flow type definitions (I wasn't able to set up the flow VS code extension in a way that it understood the typed .js files in this repo)


xtuc (approved):
thanks

PR #971, opened 2021-09-02
Account for sign in leb128 i64 decoding, fixes #970

### LINKED ISSUE #970: leb128 incorrectly decodes negative i64
The leb128 package has an error when decoding negative numbers to i64. As an example, `-1` gets decoded as `255`. Similarly, all negative integers below a certain size get decoded incorrectly.

I already identitfied the problem and will create a PR that fixes it. The reason is that the sign of the intermediate ("bigint-ish") representation returned by `encodeBufferCommon()` is not taken into account when this representation is transformed into a `Long` instance with `Long.fromBytesLE()`. This can be fixed by checking for the sign and extending the intermediate representation with `11111111` bytes at the beginning if the number is negative.

To reproduce this issue, try parsing a wasm version of the module `(module (global i64 (i64.const -1)))` with wasm-parser and observe that the AST that is produced has a `"LongNumberLiteral"`that represents `255` instead of `-1`. My PR will add a test that exposes this.

mitschabaude:
This PR

* fixes a major error in the leb128 package, described in #970 
* adds a test that would fail without the PR but succeeds with it
* fixes one other test case that checked against wasm code containing incorrectly decoded negative i64s

### DIFF
--- a/packages/leb128/src/leb.js
+++ b/packages/leb128/src/leb.js
@@ -242,6 +242,13 @@ function encodeInt64(num) {
 function decodeInt64(encodedBuffer, index) {
   const result = decodeIntBuffer(encodedBuffer, index);
 
+  // sign-extend if necessary
+  const length = result.value.length;
+  if (result.value[length - 1] >> 7) {
+    result.value = bufs.resize(result.value, 8);
+    result.value.fill(255, length);
+  }
+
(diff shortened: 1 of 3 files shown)


xtuc (approved):
Nice, good catch!

PR #1117, opened 2022-06-25
wasm-parser: skip rather than throw on unknown sections

bpowers:
One new section type is 0xd `exception`, which doesn't need any special processing, and the details aren't useful to e.g. Webpack.  It seems most robust to just skip unknown sections by default rather than have to make a change + release to unbreak parsing for spec additions.

I also added an option `errorOnUnknownSection` to keep the old behavior, in case people want to match the old behavior.

I tested this by patching a local project that uses the new exception instructions (`throw`, `try`, etc) in a WebAssembly file that is processed by Webpack and it works well.

Finally I bumped the version for `wasm-parser`.  If this looks good + mergable (happy to make changes, too!), @xtuc could you cut a release?

### DIFF
--- a/packages/wasm-parser/package.json
+++ b/packages/wasm-parser/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@webassemblyjs/wasm-parser",
-  "version": "1.11.1",
+  "version": "1.11.2",
   "keywords": [
     "webassembly",
     "javascript",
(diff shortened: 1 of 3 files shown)


xtuc (approved):
Nice thanks

xtuc:
@bpowers your change as been publish as `@1.11.3`

bpowers:
amazing, thanks @xtuc !

xtuc:
@bpowers  do you plan to update webassemblyjs in Webpack?

Issue #922, opened 2021-05-28
wasm-parser exceeds the maximum stack size on large function names in the "name" section


RSSchermer:
This problem originates from a [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen) application packaged with webpack and the [wasm-pack-plugin](https://github.com/wasm-tool/wasm-pack-plugin). [wasm-bindgen issue #2567](https://github.com/rustwasm/wasm-bindgen/issues/2567) contains details on the error I encountered.

It seems that very large function names can cause [parseNameSectionFunctions](https://github.com/xtuc/webassemblyjs/blob/0fb788b1b7785f2dcbb1b2e03b6564f607b4358c/packages/wasm-parser/src/decoder.js#L1184) to exceed the stack size. It is not clear to me how this happens. I would have expected V8 to allocate large objects on the heap. Could it be because [functionNames](https://github.com/xtuc/webassemblyjs/blob/0fb788b1b7785f2dcbb1b2e03b6564f607b4358c/packages/wasm-parser/src/decoder.js#L1185) is marked `const`?

I will concur that the function name that seems to trigger the error is comically large. However, it results from (imho) reasonable Rust code.

I can share the `pkg` directory output by wasm-bindgen if that would help (which includes the wasm binary with the offending "name" section).

Issue #918, opened 2021-05-20
Unexpected section: 0x6


phiresky:
When trying to parse the attached wasm module, webpack throws the above error.

When using the wasm-parser api directly, it shows this:

```
Failed to decode custom "name" section @5033477; ignoring (Maximum call stack size exceeded).
Uncaught Error: Unexpected section: 0x6
    at new CompileError (
[b.wasm.zip](https://github.com/xtuc/webassemblyjs/files/6516196/b.wasm.zip)
node_modules/@webassemblyjs/helper-api-error/lib/index.js:40:109)
    at parseSection (node_modules/@webassemblyjs/wasm-parser/lib/decoder.js:1329:76)
```
the file is valid according to wasm-objdump etc:

```
b.wasm: file format wasm 0x1

Sections:

     Type start=0x0000000e end=0x000002bf (size=0x000002b1) count: 100
   Import start=0x000002c5 end=0x00000352 (size=0x0000008d) count: 3
 Function start=0x00000358 end=0x00004342 (size=0x00003fea) count: 16360
    Table start=0x00004348 end=0x0000434f (size=0x00000007) count: 1
   Memory start=0x00004355 end=0x00004358 (size=0x00000003) count: 1
   Global start=0x0000435e end=0x00004367 (size=0x00000009) count: 1
   Export start=0x0000436d end=0x000043e2 (size=0x00000075) count: 6
     Elem start=0x000043e8 end=0x00004c04 (size=0x0000081c) count: 1
     Code start=0x00004c0a end=0x00407c45 (size=0x0040303b) count: 16360
     Data start=0x00407c4b end=0x0044b30c (size=0x000436c1) count: 2
   Custom start=0x0044b312 end=0x005c7e00 (size=0x0017caee) "name"
   Custom start=0x005c7e06 end=0x005c7e7d (size=0x00000077) "producers"
```

[b.wasm.zip](https://github.com/xtuc/webassemblyjs/files/6516198/b.wasm.zip)

phiresky:
The same file is also works correctly in browsers

phiresky:
replacing this: https://github.com/xtuc/webassemblyjs/blob/0fb788b1b7785f2dcbb1b2e03b6564f607b4358c/packages/wasm-parser/src/decoder.js#L1836

with ` offset = startLoc + sectionSizeInBytes` works, so seems like the issue is the offset is wrong after failing to decode the name section.

PR #1136, opened 2022-11-20
Change utf8 decoder to be iterative and not overflow the call stack

deongroenewald:
I was getting errors similar to #918 and #922 (converting a rust lib to WASM) and noted that the stack was overflowing due to the recursion in the `_decode()` function of the `utf8` package.

Changing the implementation to be iterative resolved these errors for me.


xtuc:
Nice, thanks for the change. I'll have a look.

deongroenewald:
Closing since #1140 has been merged.

PR #1140, opened 2023-04-22
Improves the utf8 decoder performance

Dig-Doug:
When parsing large wasm files, the decode function is very slow. You can reproduce the performance issues with the information in https://github.com/webpack/webpack/issues/17044. This change refactors the recursive algorithm to an iterative version that should have the exact same behavior with significantly better performance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,6 @@ esm/
 package-lock.json
 
 wasm_test_dir
+
+# IntelliJ
+.idea/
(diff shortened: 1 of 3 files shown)


xtuc:
Thanks for the PR @Dig-Doug . For the record and since the CI is broken, the tests pass locally with your changes.

xtuc (approved):

xtuc:
Released in 1.11.6.

PR #1144, opened 2023-07-26
Replace NodeJS Buffer with Uint8Array

### LINKED ISSUE #1134: use uint8array instead
just browsing the code and just having a look at it to see if it's something i would use...

something i notice right from the bat, is that you are using NodeJS specific stuff, and i'm targeting both browser and Deno, so i would prefer if you used uint8array + dataview instead.

Here is just one example...
https://github.com/xtuc/webassemblyjs/blob/255ab4de257925d7893f0ad7223fdd29c873cd44/packages/leb128/src/bufs.js#L87

don't really wish to ship node specific stuff that works slower in the browser than using say eg TextEncoder/TextDecoder etc...

wikiwong:
This PR replaces the use of the NodeJS specific `Buffer` object with a JavaScript Uint8Array so that this library may work on Deno, Node, and Web platforms without having to be polyfilled.

Closes #1134

### DIFF
--- a/packages/helper-buffer/src/index.js
+++ b/packages/helper-buffer/src/index.js
@@ -42,7 +42,7 @@ export function makeBuffer(...splitedBytes: Array<Array<Byte>>): Array<Byte> {
   return new Uint8Array(bytes).buffer;
 }
 
-export function fromHexdump(str: string): Buffer {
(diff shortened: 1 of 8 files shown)


xtuc (approved):
nice! thanks for your work.

PR #1147, opened 2024-01-16
add narrow-width integer sign bytes for helper-wasm-bytecode

pi0:
reference: https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#narrow-width-integer-sign-extension


xtuc (approved):
nice, thanks!

